Issue #15, opened 2022-07-30
When the server is unavailable, the client does not have any exception logs.


rxf113:
### Version

4.3.1 and 4.3.3-SNAPSHOT

### Context

My server is interrupted while running, the client does not have any exceptions or logs, It is not easy to locate the problem in time.
I looked at the source code and found that in the `vertx-grpc-client` module `io.vertx.grpc.client.VertxClientCall #start()` does not handle the client.request error.

PR #16, opened 2022-07-30
When client.request failed, throw the exception

rxf113:
fix bug #15


tsegismont:
cc @vietj

vietj:
Can you add a test ?

rxf113:
> Can you add a test ?

I added a test, and There are some other changes, please confirm carefully

rxf113:
> > Can you add a test ?
> 
> I added a test, and There are some other changes, please confirm carefully

Added a test sample  [vertx-grpc-server-unavailable](https://github.com/rxf113/vertx-grpc-server-unavailable)

vietj:
the test has been removed, can you elaborate ?

rxf113:
> the test has been removed, can you elaborate ?
I added a test sample  [vertx-grpc-server-unavailable](https://github.com/rxf113/vertx-grpc-server-unavailable)

The previous test shouldn't be there because it needs to depend on VetxStub and I had to modify the pom.xml, which I don't think is a qualifying unit test, so i removed it

rxf113:
Updated description：  https://github.com/rxf113/vertx-grpc-server-unavailable/blob/main/README.md

liangyuanpeng (commented):
Maybe add test case on this repo would be great.

vertx-grpc-client/src/main/java/io/vertx/grpc/client/VertxClientCall.java
```diff
@@ -121,6 +123,12 @@ public void start(Listener<ResponseT> responseListener, Metadata headers) {
           }
         });
         writeAdapter.init(request, new BridgeMessageEncoder<>(methodDescriptor.getRequestMarshaller(), compressor));
+      } else {
+        //Maybe the client.request will fail, eg: server interrupt
+        Throwable err = ar1.cause();
+        if (err instanceof ConnectException || err instanceof ClosedChannelException){
```
vietj, line 129: I think we already have error handling in the second asynchronous result that performs a close operation, we should try merging the two in a `compose` operation and handle a single doClose in the composed future.

vietj:
thanks, I ported the code here in the more general PR about vertx client call error handling : https://github.com/eclipse-vertx/vertx-grpc/pull/19

PR #31, opened 2023-02-01
gRPC server will send header/trailer frames instead of a single frame when a trailers only case happens

vietj:
gRPC mandates to send a single response frame when an immediate occurs, the vertx server does not support it when a response trailer metadata is set, instead it will send a header frame and a trailer frame.


vietj:
cc @cescoffier @alesj

cescoffier (commented):
LGTM

PR #36, opened 2023-02-15
Add a shortcut method to get grpc reply

liuchong:
Motivation:

In many usage cases we will call end and last, give a default shortcut method for this should be convenient I think 👀 

Please review when you have time @vietj @tsegismont , thanks!


vietj:
I like the idea but I think it is too narrow and specific, I created a PR following your idea https://github.com/eclipse-vertx/vertx-grpc/pull/38, please have a look

liuchong:
I've reviewed your PR and it looks nice!
Also maybe this `last` is still somewhat useful in some certain use cases, as a `shortcut`, how do you think? 🤔

vietj:
I think that's too narrow to be here

liuchong:
ok

PR #38, opened 2023-02-15
GrpcClient API simplified call API to simplify interactions with a gRPC service 

vietj:
A simplified API to interact with a gRPC service:

```java
Future<T> fut = client.call(server, service, req -> req.send(item), resp -> resp.last());
```

```java
Future<List<T>> fut = client.call(server, service, req -> req.send(item), resp -> resp.collecting(Collectors.toList()));
```


liuchong (approved):
Great! 👍

PR #48, opened 2023-03-10
New feature: vertx grpc protoc plugin

lwlee2608:
Create a protoc-plugin that uses the latest vertx-grpc-server and vertx-grpc-client library. This plugin enforces the use of RPCs that are only defined in the proto specification, ensuring greater consistency in communication between the server and client. Additionally, the generated code provides a clearer interface for the user and reduces boilerplate code

PR #50, opened 2023-03-14
New feature: vertx grpc protoc plugin

lwlee2608:
Create a protoc-plugin that uses the latest vertx-grpc-server and vertx-grpc-client library. This plugin enforces the use of RPCs that are only defined in the proto specification, ensuring greater consistency in communication between the server and client. Additionally, the generated code provides a clearer interface for the user and reduces boilerplate code


lwlee2608:
duplicate with https://github.com/eclipse-vertx/vertx-grpc/pull/48

Issue #56, opened 2023-05-27
IndexOutOfBoundsException in GrpcMethodCall


SG-O:
The method serviceName() in GrpcMethodCall can throw an index out of bounds exception if the service path doesn't contain a dot.

I encountered this bug in 4.4.2.

The path I tried it with is "/Auth/GetChallenge"

PR #57, opened 2023-05-27
Fixed IndexOutOfBoundsException in GrpcMethodCall.serviceName()

SG-O:
This should fix the issue https://github.com/eclipse-vertx/vertx-grpc/issues/56

### DIFF
--- a/vertx-grpc-common/src/main/java/io/vertx/grpc/common/impl/GrpcMethodCall.java
+++ b/vertx-grpc-common/src/main/java/io/vertx/grpc/common/impl/GrpcMethodCall.java
@@ -46,7 +46,7 @@ public class GrpcMethodCall {
   public String methodName() {
     if (methodName == null) {
       int idx = path.lastIndexOf('/');
-      methodName = path.substring(idx);
+      methodName = path.substring(idx + 1);
     }
     return methodName;
   }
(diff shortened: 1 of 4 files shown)


vietj:
can you add a test @SG-O ?

SG-O:
I added test for the whole GrpcMethodCall class. 

Implementing these revealed that the output of the fullyQualifiedName() method in ServiceNameImpl wasn't formatted well for service names without dots. I fixed this corner case so it now outputs "MyService" instead of ".MyService".

SG-O:
I found another two minor bugs. 

In the implementation of GrpcMethodCall.methodName() the name would have a leading slash (e.g. /Method1) which is not compatible with ServiceName.pathOf(String method). I changed the output to not include this slash.

The ServiceName.packageName() and .pathOf() methods did not handle fully qualified names without a package name correctly. I fixed this and added a lot more tests for the whole ServiceName class.

Both GrpcMethodCall and ServiceName now have full coverage from their respective tests.

This should be ready to  merge.

vietj:
can you resolved the conflict due to merging https://github.com/eclipse-vertx/vertx-grpc/pull/59 ? @SG-O

SG-O:
On it

vietj:
can you provide a backport to 4.x branch @SG-O ?

PR #63, opened 2023-06-02
Vertx grpc protoc plugin

lwlee2608:
Create a protoc-plugin that uses the latest vertx-grpc-server and vertx-grpc-client library. This plugin enforces the use of RPCs that are only defined in the proto specification, ensuring greater consistency in communication between the server and client. Additionally, the generated code provides a clearer interface for the user and reduces boilerplate code

backport PR #48


vietj:
thanks @lwlee2608

Issue #37, opened 2023-02-15
Support Timeout Propagation


coding4m:
As title.
#### Refer issue: 
[Context deadline propagation should cascade](https://github.com/grpc/grpc-java/issues/1205#issue-116811766)

PR #64, opened 2023-06-12
Add GrpcClientRequestOptions to pass timeout

liuchong:
Motivation:

I want to pass timeout option when request grpc server with client, so added a new GrpcClientReqesatOptions class to impl it.
In future maybe we could add more options into it from io.vertx.core.http.RequestOptions

### DIFF
--- a/vertx-grpc-client/src/main/generated/io/vertx/grpc/client/GrpcClientRequestOptionsConverter.java
+++ b/vertx-grpc-client/src/main/generated/io/vertx/grpc/client/GrpcClientRequestOptionsConverter.java
@@ -1,0 +1,39 @@
+package io.vertx.grpc.client;
+
+import io.vertx.core.json.JsonObject;
+import io.vertx.core.json.JsonArray;
+import io.vertx.core.json.impl.JsonUtil;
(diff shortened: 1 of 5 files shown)


liuchong:
Could you please check that if we need this when you have time? @vietj @tsegismont 👀

vietj:
I'll have a look

vietj:
can you add a test for this @liuchong ?

liuchong:
ok I will do 👀

liangyuanpeng (commented):
It would be resolve the issue #37  , right?

liangyuanpeng:
> It would be resolve the issue https://github.com/eclipse-vertx/vertx-grpc/issues/37 , right?

It's not true, #37 is talking ablout server side and this PR is work for client.

vietj:
see the idle timeout and the deadline & timeout PR

liuchong:
https://github.com/eclipse-vertx/vertx-grpc/pull/78
ok, seems the similar logic we've talked about it.

liuchong:
ok, I fount then in #79 and #78, seems using the logic we've talked before, thanks! 👀

PR #68, opened 2023-08-22
Avoid composite buffers for small messages

franz1981:
This change is optimizing the use case of Netty composite buffers in case a small payload/data is required.

The NIO Netty flow will end up copying the compontents from the CompositeBuffer in a single native direct buffer at 
https://github.com/netty/netty/blob/fbb0207d5ecce39f3d63450dfd59bad5510b8e8b/transport/src/main/java/io/netty/channel/nio/AbstractNioChannel.java#L443.

For a composite buffer made of 2 components (the length and compression flag + data ones) it means iterating them (updating for each the offsets etc etc) in `CompositeByteBuf::getBytes(int index, ByteBuf dst, int dstIndex, int length)` (calling back to `UnpooledHeapByteBuf::getBytes(int index, ByteBuf dst, int dstIndex, int length)` with the direct `dst` for each component, updating each ones offset, checking accessibility etc etc).

In the case of a single merged buffer, we pay an additional allocation (and copy of data), but:
- the allocation is actually amortized: (for JDK 17 and COOPS/CCPS) allocating the composite + the length buffer still cost a total of `+176 bytes`, while allocating a new buffer (not composite) would cost `data length + 5 + 48 bytes` and perform the additional data copy. eg for 128 bytes data, the composite would still allocate +176 bytes, while this PR will allocate 128 + 53 = 181 bytes which is similar (but we can throw away immediately the original data buffer, no longer needed!)
- no ping-pong of buffer types to find the appropriate copy method nor iterations required (or accessiblity/offsets adjustments), just calling straight to `setBytes(AbstractByteBuf buf, long addr, int index, ByteBuf src, int srcIndex, int length)` on the direct pooled buffer, using a single vertx heap buffer which contains length + data.  

The point about reachability of buffers is a stealthy but important consideration:
- in the original code version the reachable live data must be the original buffer AND the composite one, till the NIO channel copy them into the direct single buffer (to be sent on the wire) eg 128 bytes original buffer + 176 composite = 304 bytes live data
- in the new code, instead, instead of the composite, we create a single new buffer, and the previous one case be GCed, making it, for 128 bytes original data, just 181 bytes in total (which is way less too!)

### DIFF
--- a/vertx-grpc-common/src/main/java/io/vertx/grpc/common/impl/GrpcMessageImpl.java
+++ b/vertx-grpc-common/src/main/java/io/vertx/grpc/common/impl/GrpcMessageImpl.java
@@ -38,15 +38,26 @@ public class GrpcMessageImpl implements GrpcMessage {
   }
 
   public static Buffer encode(GrpcMessage message) {
-    ByteBuf bbuf = ((BufferInternal)message.payload()).getByteBuf();
+    ByteBuf bbuf = ((BufferInternal) message.payload()).getByteBuf();
     int len = bbuf.readableBytes();
     boolean compressed = !message.encoding().equals("identity");
+    if (len <= 128) {
+      // it is worthy to just copy here, 'cause composite (heap) buffers are slower to be sent to the wire or copied
+      int totalBytes = 5 + len;
+      ByteBuf fullMsg = Unpooled.buffer(totalBytes, totalBytes);
+      fullMsg.setByte(0, compressed ? 1 : 0); // Compression flag
+      fullMsg.setInt(1, len);                 // Length
+      fullMsg.setBytes(5, bbuf, bbuf.readerIndex(), len);
+      fullMsg.writerIndex(totalBytes);
+      return BufferInternal.buffer(fullMsg);
+    }
+    // slow-path
     ByteBuf prefix = Unpooled.buffer(5, 5);
     prefix.writeByte(compressed ? 1 : 0);      // Compression flag
     prefix.writeInt(len);                      // Length
-    CompositeByteBuf composite = Unpooled.compositeBuffer();
-    composite.addComponent(true, prefix);
-    composite.addComponent(true, bbuf);
+    CompositeByteBuf composite = Unpooled.compositeBuffer(2);
+    composite.addComponent(true, 0, prefix);
+    composite.addComponent(true, 1, bbuf);
     return BufferInternal.buffer(composite);
   }
 }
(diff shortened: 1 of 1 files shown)


vertx-grpc-common/src/main/java/io/vertx/grpc/common/impl/GrpcMessageImpl.java
```diff
@@ -38,15 +39,31 @@ public Buffer payload() {
...
+      // let's use vertx buffer here because it doesn't have any atomic release, if unpooled
+      ByteBuf fullMsg = VertxByteBufAllocator.DEFAULT.heapBuffer(totalBytes, totalBytes);
+      fullMsg.setByte(0, compressed ? 1 : 0); // Compression flag
+      fullMsg.setInt(1, len);                 // Length
+      fullMsg.setBytes(5, bbuf, bbuf.readerIndex(), len);
+      fullMsg.writerIndex(totalBytes);
+      try {
+        return BufferInternal.buffer(fullMsg);
+      } finally {
+        bbuf.release();
```
vietj, line 57: vertx buffers are unreleasable and not pooled

franz1981:
@He-Pin I remember you have commented on the gRpc benchmark...we are still working on enabling our load generator to support gRpc, hence we have no "official" and supported benchmarks to verify this. This is mostly based on my knowledge of the Netty stack and the OpenJDK platform; do you have anything to test it and verify if it is worthy?

vertx-grpc-common/src/main/java/io/vertx/grpc/common/impl/GrpcMessageImpl.java
```diff
@@ -38,15 +39,31 @@ public Buffer payload() {
...
   public static Buffer encode(GrpcMessage message) {
-    ByteBuf bbuf = ((BufferInternal)message.payload()).getByteBuf();
+    ByteBuf bbuf = ((BufferInternal) message.payload()).getByteBuf();
     int len = bbuf.readableBytes();
     boolean compressed = !message.encoding().equals("identity");
+    // it is worthy to just copy here, 'cause composite (heap) buffers are slower to be sent to the wire or copied
+    if (len <= 128) {
+      int totalBytes = 5 + len;
+      // let's use vertx buffer here because it doesn't have any atomic release, if unpooled
+      ByteBuf fullMsg = VertxByteBufAllocator.DEFAULT.heapBuffer(totalBytes, totalBytes);
```
franz1981, line 49: Maybe we would like to use pooled direct buffers here instead, wdyt @vietj ?
It would save the transport later to copy it into a direct one :/

PR #73, opened 2023-10-12
[4.x] update javax to Jakarta

liangyuanpeng:
Motivation:

Explain here the context, and why you're making that change, what is the problem you're trying to solve.

Upgrade to jakarta-annotation-api, javax.annotation-api has been deprecated

related: https://github.com/etcd-io/jetcd/issues/1243

Conformance:

### DIFF
--- a/vertx-grpc-client/src/main/java/examples/GreeterGrpc.java
+++ b/vertx-grpc-client/src/main/java/examples/GreeterGrpc.java
@@ -7,7 +7,7 @@ import static io.grpc.MethodDescriptor.generateFullMethodName;
  * The greeting service definition.
  * </pre>
  */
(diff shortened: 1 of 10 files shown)


liangyuanpeng:
May related: 
- https://github.com/grpc/grpc-java/issues/9179

vertx-grpc-client/src/main/java/examples/GreeterGrpc.java
```diff
@@ -7,7 +7,7 @@
  * The greeting service definition.
  * </pre>
  */
-@javax.annotation.Generated(
```
vietj, line 10: it is not clear why this is deprecated as this is part of the JDK

PR #75, opened 2023-10-30
Add JWT support for gRPC server

Jotschi:
Motivation:

This PR adds JWT support to the GrpcServer.

TODO: 
* [x] Add usage doc


vietj:
I think we need to go with a builder for the `GrpcServer` to add authentication handler instead of overloading the `GrpcServer` factory methods

```java
interface GrpcServerBuilder {
  GrpcServerBuilder withAuthHandler(...);
}
```

vietj:
Shouldn't we set an auth handler when we do register a handler instead of being global to the server ?

Jotschi:
> Shouldn't we set an auth handler when we do register a handler instead of being global to the server ?

Right. That would be better. I changed the API.

Jotschi:
@vietj 
I have squashed the commits of this PR and cherry-picked it into 4.x because I need this feature for my Vert.x 4.5.x project.  This is just a headsup to let you know that this PR would also work for 4.x

vietj:
can we also expect Oauth2 support and client support ?

vertx-grpc-server/src/main/java/io/vertx/grpc/server/GrpcServerRequest.java
```diff
@@ -78,4 +79,9 @@ public interface GrpcServerRequest<Req, Resp> extends GrpcReadStream<Req> {
    * @return the underlying HTTP connection
    */
   HttpConnection connection();
+
+  User user();
+
+  GrpcServerRequest<Req, Resp> setUser(User user);
```
vietj, line 85: we don't need this I think

Jotschi: I removed the setter from the interface but the getter is required. Otherwise a user of the API is unable to access the authenticated user. At this point there is no "Context" object for gRPC so I put the user in the request.

Jotschi:
> can we also expect Oauth2 support and client support ?

I added token credential support to the client. As for Oauth2 support I don't feel comfortable to add it at this stage.

vietj:
can you describe the issue for oauth2 ? it seems pretty ok to me, but pehraps I'm missing something

Jotschi:
@vietj I took a look at the existing webclient implementation https://github.com/vert-x3/vertx-web/blob/master/vertx-web-client/src/main/java/io/vertx/ext/web/client/impl/OAuth2AwareInterceptor.java
I assume it would need to work in a similar way and basically intercept the calls and check whether the token needs to be refreshed. For this PR I would like to focus on JWT. Adding OAuth2 would mean creating an additional maven module. (eg. oauth2-grpc-client). I'm however getting a bit time constrained at the moment and can't add it at this point of time.

vertx-grpc-client/src/main/java/io/vertx/grpc/client/GrpcClient.java
```diff
@@ -112,6 +112,15 @@ default <Req, Resp, T> Future<T> call(SocketAddress server, MethodDescriptor<Req
...
   }
 
+  /**
+   * Mark that request should be dispatched with authentication obtained from passed {@code JWTAuth} provider
+   *
+   * @param credentials
+   * @return a reference to this, so the API can be used fluently
+   */
+  @GenIgnore(GenIgnore.PERMITTED_TYPE)
+  GrpcClient withCredentials(Credentials credentials);
```
vietj, line 122: This should be named `credentials`, we prefer to use `withXXX` in builders

vertx-grpc-server/src/main/java/io/vertx/grpc/server/impl/GrpcServerImpl.java
```diff
@@ -59,6 +65,27 @@ public void handle(HttpServerRequest httpRequest) {
   private <Req, Resp> void handle(MethodCallHandler<Req, Resp> method, HttpServerRequest httpRequest, GrpcMethodCall methodCall) {
     GrpcServerRequestImpl<Req, Resp> grpcRequest = new GrpcServerRequestImpl<>(httpRequest, method.messageDecoder, method.messageEncoder, methodCall);
     grpcRequest.init();
+    GrpcAuthenticationHandler authHandler = method.authHandler;
+    if (authHandler != null) {
```
vietj, line 69: here we need to pause the grpc request and resume it after to avoid loosing messages

vertx-grpc-server/src/main/java/io/vertx/grpc/server/auth/impl/GrpcJWTAuthenticationHandlerImpl.java
```diff
@@ -0,0 +1,137 @@
...
+        .replaceAll("\"", "\\\"");
+
+    if (this.realm != null &&
+      (this.realm.indexOf('\r') != -1 || this.realm.indexOf('\n') != -1)) {
+      throw new IllegalArgumentException(
+        "Not allowed [\\r|\\n] characters detected on realm name");
+    }
+  }
+
+  private Future<String> parseAuthorization(HttpServerRequest req,
```
vietj, line 68: we should use a result here and avoid relying on exception to communicate an authentication failure (and thus avoid GrpcException), the result should clearly be an enum like result and the gRPC code calling for authentication should set the corresponding status on the HTTP/2 response

vietj (changes requested):
thanks for the contribution and sorry for the late review.

vertx-grpc-server/src/main/java/io/vertx/grpc/server/impl/GrpcServerImpl.java
Jotschi: Is this required because in `ìnit` additional handlers are being added?

vertx-grpc-server/src/main/java/io/vertx/grpc/server/auth/impl/GrpcJWTAuthenticationHandlerImpl.java
Jotschi: I agree. Failures should only be used for unexpected states. I started rewriting this to use an enum as suggested:

```java
     return Future.succeededFuture(GrpcAuthenticationResult.UNAUTHENTICATED);
```

I however don't see how I can return the token string in the same method via:

```java
    return Future.succeededFuture(authorization.substring(idx + 1));
```

Do have an example or a pointer for me on how to solve this?

vertx-grpc-client/src/main/java/io/vertx/grpc/client/GrpcClient.java
Jotschi: Updated

vertx-grpc-server/src/main/java/io/vertx/grpc/server/auth/impl/GrpcJWTAuthenticationHandlerImpl.java
vietj: I will need to have a look deeper to make an elaborated suggestion here

vertx-grpc-server/src/main/java/io/vertx/grpc/server/impl/GrpcServerImpl.java
vietj: I think you can write a test for this, by sending messages in the client and have an authentication handler that delays the actual auth handler by a delay and check we got all messages on the server

Issue #84, opened 2023-12-22
GrpcServer can't be mounted in a Vert.x Web router 


tsegismont:
The [documentation](https://github.com/eclipse-vertx/vertx-grpc/blame/main/vertx-grpc-server/src/main/asciidoc/server.adoc#L42) and the [4.3 release blog post](https://vertx.io/blog/whats-new-in-vert-x-4-3/#vertx-grpc) indicate that a `GrpcServer` can be mounted in a Vert.x Web router.

In fact, it is not possible because the gRPC server is a `Handler<HttpServerRequest>` not a `Handler<RoutingContext>`. So this blog post example is wrong:

```java
GreeterGrpc.GreeterImplBase service = new GreeterGrpc.GreeterImplBase() {
  @Override
  public void sayHello(HelloRequest request, StreamObserver<HelloReply> responseObserver) {
    responseObserver.onNext(HelloReply.newBuilder().setMessage("Hello " + request.getName()).build());
    responseObserver.onCompleted();
  }
};

GrpcServer grpcServer = GrpcServer.server(vertx);
GrpcServiceBridge serverStub = GrpcServiceBridge.bridge(service);
serverStub.bind(grpcServer);

router.consumes("application/grpc").handler(grpcServer); // does not compile
```

tsegismont:
@vietj did I miss anything?

vietj:
I would like to avoid this dependency of vertx-grpc over vertx-web for various reasons.

Instead we should provide the code `return rc -> server.handle(rc.request())` in the documentation, alternatively we could consider having a dependency of vertx-web onto vertx-grpc, but I think that is not necessary.

tsegismont:
OK, let's fix the documentation and the blog post

tsegismont:
Fixed by 31d21d8b9709e37ab67c78d03ee48db8c9ed12bf

PR #86, opened 2024-01-03
Add GrpcServerRouteHandler

tsegismont:
See #84

In order to integrate Vert.x Web routes and gRPC servers, we need an adapter.


vietj:
I would like to avoid this dependency of vertx-grpc over vertx-web for various reasons.

Instead we should provide the code `return rc -> server.handle(rc.request())` in the documentation, alternatively we could consider having a dependency of vertx-web onto vertx-grpc, but I think that is not necessary.

PR #88, opened 2024-02-12
HTTP client updates

vietj:
The gRPC client builds its own HTTP client. Since the HTTP client got new capabilities (service name resolution, unpooled connections), it makes sense to allow wrapping an existing HTTP client interface.

Modify the gRPC client to allow wrapping an existing HTTP client interface.


vietj:
FYI @tsegismont this updates the gRPC client to allow using an `HttpClient` since the `HttpClientConnection` PR has been merged

PR #89, opened 2024-02-21
Implement gRPC-Web protocol

### LINKED ISSUE #55: gRPC Web
Support [grpc-web protocol](https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-WEB.md) in Vert.x gRPC Server

tsegismont:
Closes #55 

Each commit can be reviewed individually.

- [X] Test suite
- [X] Implementation
- [x] Documentation
- [x] Integration test with Vert.x Web, CORS and a browser app ([binary](https://github.com/tsegismont/grpc-web-hello-world/tree/with-vertx-grpc-web) and [text](https://github.com/tsegismont/grpc-web-hello-world/tree/with-vertx-grpc-web-text))
- [x] grpc-web interop tests and automation

### DIFF
--- a/vertx-grpc-common/src/main/java/io/vertx/grpc/common/GrpcMediaType.java
+++ b/vertx-grpc-common/src/main/java/io/vertx/grpc/common/GrpcMediaType.java
@@ -1,0 +1,74 @@
+/*
+ * Copyright (c) 2011-2024 Contributors to the Eclipse Foundation
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0, or the Apache License, Version 2.0
+ * which is available at https://www.apache.org/licenses/LICENSE-2.0.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR Apache-2.0
+ */
+
+package io.vertx.grpc.common;
+
+import io.netty.util.AsciiString;
+import io.vertx.codegen.annotations.Unstable;
+import io.vertx.core.http.HttpHeaders;
+
+/**
+ * The gRPC media types.
+ */
+@Unstable
+public class GrpcMediaType {
+
+  /**
+   * gRPC.
+   */
+  public static final CharSequence GRPC = HttpHeaders.createOptimized("application/grpc");
+  /**
+   * gRPC with Protobuf message format.
+   */
+  public static final CharSequence GRPC_PROTO = HttpHeaders.createOptimized("application/grpc+proto");
+
+  /**
+   * gRPC Web binary.
+   */
+  public static final CharSequence GRPC_WEB = HttpHeaders.createOptimized("application/grpc-web");
+  /**
+   * gRPC Web binary with Protobuf message format.
+   */
+  public static final CharSequence GRPC_WEB_PROTO = HttpHeaders.createOptimized("application/grpc-web+proto");
+
+  /**
+   * Whether the provided {@code mediaType} represents gRPC-Web
+   *
+   * @param mediaType the value to test
+   * @return {@code true} if the value represents gRPC-Web, {@code false} otherwise
+   */
+  public static boolean isGrpcWeb(CharSequence mediaType) {
+    return AsciiString.regionMatches(GRPC_WEB, true, 0, mediaType, 0, GRPC_WEB.length());
+  }
+
+  /**
+   * gRPC Web text (base64).
+   */
+  public static final CharSequence GRPC_WEB_TEXT = HttpHeaders.createOptimized("application/grpc-web-text");
+  /**
+   * gRPC Web text (base64) with Protobuf message format.
+   */
+  public static final CharSequence GRPC_WEB_TEXT_PROTO = HttpHeaders.createOptimized("application/grpc-web-text+proto");
+
+  /**
+   * Whether the provided {@code mediaType} represents gRPC-Web
+   *
+   * @param mediaType the value to test
+   * @return {@code true} if the value represents gRPC-Web, {@code false} otherwise
+   */
+  public static boolean isGrpcWebText(CharSequence mediaType) {
+    return AsciiString.regionMatches(GRPC_WEB_TEXT, true, 0, mediaType, 0, GRPC_WEB_TEXT.length());
+  }
+
+  private GrpcMediaType() {
+    // Constants
+  }
+}
--- a/vertx-grpc-common/src/main/java/io/vertx/grpc/common/GrpcMediaType.java
+++ b/vertx-grpc-common/src/main/java/io/vertx/grpc/common/GrpcMediaType.java
@@ -19,7 +19,7 @@ import io.vertx.core.http.HttpHeaders;
  * The gRPC media types.
  */
 @Unstable
-public class GrpcMediaType {
+public final class GrpcMediaType {
 
   /**
    * gRPC.
--- a/vertx-grpc-common/src/main/java/io/vertx/grpc/common/impl/GrpcMessageImpl.java
+++ b/vertx-grpc-common/src/main/java/io/vertx/grpc/common/impl/GrpcMessageImpl.java
@@ -21,10 +21,18 @@ public class GrpcMessageImpl implements GrpcMessage {
 
   private final String encoding;
   private final Buffer payload;
+  private final boolean trailer;
 
   public GrpcMessageImpl(String encoding, Buffer payload) {
     this.encoding = encoding;
     this.payload = payload;
+    trailer = false;
+  }
+
+  public GrpcMessageImpl(String encoding, Buffer payload, boolean trailer) {
+    this.encoding = encoding;
+    this.payload = payload;
+    this.trailer = trailer;
   }
 
   @Override
@@ -41,8 +49,9 @@ public class GrpcMessageImpl implements GrpcMessage {
     ByteBuf bbuf = ((BufferInternal)message.payload()).getByteBuf();
     int len = bbuf.readableBytes();
     boolean compressed = !message.encoding().equals("identity");
+    boolean trailer = message instanceof GrpcMessageImpl && ((GrpcMessageImpl) message).trailer;
     ByteBuf prefix = Unpooled.buffer(5, 5);
-    prefix.writeByte(compressed ? 1 : 0);      // Compression flag
+    prefix.writeByte((trailer ? 0x80 : 0x00) | (compressed ? 0x01 : 0x00));
     prefix.writeInt(len);                      // Length
     CompositeByteBuf composite = Unpooled.compositeBuffer();
     composite.addComponent(true, prefix);
--- a/vertx-grpc-common/src/main/java/io/vertx/grpc/common/impl/GrpcMessageImpl.java
+++ b/vertx-grpc-common/src/main/java/io/vertx/grpc/common/impl/GrpcMessageImpl.java
@@ -41,6 +41,13 @@ public class GrpcMessageImpl implements GrpcMessage {
     return encode(message, false);
   }
 
+  /**
+   * Encode a {@link GrpcMessage}.
+   *
+   * @param message the message
+   * @param trailer whether this message is a gRPC-Web trailer
+   * @return the encoded message
+   */
   public static BufferInternal encode(GrpcMessage message, boolean trailer) {
     ByteBuf bbuf = ((BufferInternal)message.payload()).getByteBuf();
     int len = bbuf.readableBytes();
--- a/vertx-grpc-common/src/main/java/io/vertx/grpc/common/impl/GrpcMessageImpl.java
+++ b/vertx-grpc-common/src/main/java/io/vertx/grpc/common/impl/GrpcMessageImpl.java
@@ -21,18 +21,10 @@ public class GrpcMessageImpl implements GrpcMessage {
 
   private final String encoding;
   private final Buffer payload;
-  private final boolean trailer;
 
   public GrpcMessageImpl(String encoding, Buffer payload) {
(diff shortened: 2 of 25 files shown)


vertx-grpc-server/src/main/java/io/vertx/grpc/server/impl/GrpcServerRequestImpl.java
```diff
@@ -91,4 +104,54 @@ public GrpcServerResponse<Req, Resp> response() {
...
+      super.handle(decoded);
+    } else if (maxBase64 > 0) {
+      Buffer decoded = Buffer.buffer(DECODER.decode(grpcWebTextBuffer.getBytes(0, maxBase64)));
+      grpcWebTextBuffer = grpcWebTextBuffer.getBuffer(maxBase64, len);
+      super.handle(decoded);
+    }
+  }
+
+  private int maxBase64(int len) {
+    while (!divisibleByFour(len)) {
```
vietj, line 152: isn't that `len & 0x3` ?

tsegismont: Yes, that's what the `divisibleByFour` method  above does. It's a small private method, its content should be easily inlined by the JVM.

`divisibleByFour` is used by another method as well, and it makes the code more readable.

tsegismont:
@vietj I've removed the milestone and label, they are already set on the corresponding issue

vertx-grpc-server/src/main/java/io/vertx/grpc/server/impl/GrpcServerRequestImpl.java
vietj, line 152: I meant that the entire `while` block is equivalent to do `len & ~0x03` which is way more efficient

tsegismont: Got you, thanks for pointing this out

tsegismont:
@vietj PTAL

tsegismont:
@vietj I've squashed commits to ease the next review

vertx-grpc-server/src/main/java/io/vertx/grpc/server/impl/GrpcServerImpl.java
```diff
@@ -15,30 +15,43 @@
...
+  private static final Logger log = LoggerFactory.getLogger(GrpcServer.class);
+
+  private final GrpcServerOptions options;
   private Handler<GrpcServerRequest<Buffer, Buffer>> requestHandler;
   private Map<String, MethodCallHandler<?, ?>> methodCallHandlers = new HashMap<>();
 
-  public GrpcServerImpl(Vertx vertx) {
-    this.vertx = vertx;
+  public GrpcServerImpl(Vertx vertx, GrpcServerOptions options) {
+    this.options = Objects.requireNonNull(options, "options is null");
```
vietj, line 47: we should copy options here

vertx-grpc-server/src/main/java/io/vertx/grpc/server/impl/GrpcServerImpl.java
```diff
@@ -56,6 +69,26 @@ public void handle(HttpServerRequest httpRequest) {
     }
   }
 
+  private boolean refuseRequest(HttpServerRequest request) {
+    if (request.version() != HttpVersion.HTTP_2) {
```
vietj, line 73: simplify by returning a response code in various checks. When this code is not -1 then the caller sends the response, otherwise it proceeds to the call.

vertx-grpc-server/src/main/java/io/vertx/grpc/server/GrpcServerOptions.java
```diff
@@ -0,0 +1,100 @@
...
+  public GrpcServerOptions setGrpcWebEnabled(boolean grpcWebEnabled) {
+    this.grpcWebEnabled = grpcWebEnabled;
+    return this;
+  }
+
+  /**
+   * Disable the gRPC-Web protocol support.
+   */
+  @GenIgnore(PERMITTED_TYPE)
+  public GrpcServerOptions disableGrpcWeb() {
```
vietj, line 80: please don't add this, it breaks the data object idioms

vertx-grpc-common/src/main/java/io/vertx/grpc/common/impl/GrpcMessageImpl.java
```diff
@@ -21,10 +21,18 @@ public class GrpcMessageImpl implements GrpcMessage {
 
   private final String encoding;
   private final Buffer payload;
+  private final boolean trailer;
```
vietj, line 24: overall this trailer field should not be present here, instead the encode message should have a parameter to specify the message should be encoded as a trailer for a given message

vertx-grpc-server/src/main/java/io/vertx/grpc/server/impl/GrpcServerResponseImpl.java
```diff
@@ -234,12 +250,33 @@ private Future<Void> writeMessage(GrpcMessage message, boolean end) {
...
+      return httpResponse.end();
     } else {
-      return httpResponse.write(GrpcMessageImpl.encode(message));
+      return httpResponse.write(encodeMessage(message));
+    }
+  }
+
+  private Buffer encodeMessage(GrpcMessage message) {
+    Buffer buffer = GrpcMessageImpl.encode(message);
+    if (contentType == GrpcMediaType.GRPC_WEB_TEXT_PROTO) {
```
vietj, line 273: I know content type is set to a constant, but I believe it is safer here to use equals than ==

vertx-grpc-server/src/main/java/io/vertx/grpc/server/impl/GrpcServerResponseImpl.java
```diff
@@ -234,12 +250,33 @@ private Future<Void> writeMessage(GrpcMessage message, boolean end) {
...
+      return Buffer.buffer(Base64.getEncoder().encode(buffer.getBytes()));
```
vietj, line 274: we should use io.netty.handler.codec.base64.Base64 here that encodes a ByteBuf to a ByteBuf which avoids ByteBuf -> byte[] -> byte[] -> ByteBuf

vertx-grpc-server/src/main/java/io/vertx/grpc/server/impl/GrpcServerRequestImpl.java
```diff
@@ -91,4 +104,44 @@ public GrpcServerResponse<Req, Resp> response() {
...
+
+  private void bufferAndDecode(Buffer chunk) {
+    grpcWebTextBuffer.appendBuffer(chunk);
+    int len = grpcWebTextBuffer.length();
+    // Decode base64 content as soon as we have more bytes than a multiple of four.
+    // We could instead wait for the buffer length to be a multiple of four,
+    // But then in the worst case we may have to buffer the whole request.
+    int maxDecodable = len & ~0b11;
+    if (maxDecodable == len) {
+      Buffer decoded = Buffer.buffer(DECODER.decode(grpcWebTextBuffer.getBytes()));
```
vietj, line 138: we should use io.netty.handler.codec.base64.Base64 here that decodes a ByteBuf to a ByteBuf which avoids ByteBuf -> byte[] -> byte[] -> ByteBuf

vietj (changes requested):
A few comments to address

vertx-grpc-common/src/main/java/io/vertx/grpc/common/impl/GrpcMessageImpl.java
```diff
@@ -41,8 +49,9 @@ public static Buffer encode(GrpcMessage message) {
     ByteBuf bbuf = ((BufferInternal)message.payload()).getByteBuf();
     int len = bbuf.readableBytes();
     boolean compressed = !message.encoding().equals("identity");
+    boolean trailer = message instanceof GrpcMessageImpl && ((GrpcMessageImpl) message).trailer;
     ByteBuf prefix = Unpooled.buffer(5, 5);
-    prefix.writeByte(compressed ? 1 : 0);      // Compression flag
+    prefix.writeByte((trailer ? 0x80 : 0x00) | (compressed ? 0x01 : 0x00));
```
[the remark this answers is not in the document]
tsegismont: Right, it's specific to gRPC-Web

vertx-grpc-common/src/main/java/io/vertx/grpc/common/GrpcMediaType.java
```diff
@@ -0,0 +1,74 @@
...
+
+import io.netty.util.AsciiString;
+import io.vertx.codegen.annotations.Unstable;
+import io.vertx.core.http.HttpHeaders;
+
+/**
+ * The gRPC media types.
+ */
+@Unstable
+public class GrpcMediaType {
```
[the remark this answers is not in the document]
tsegismont: It's effectively final (private constructor).

vertx-grpc-server/src/main/java/io/vertx/grpc/server/impl/GrpcServerImpl.java
tsegismont: ok

vertx-grpc-server/src/main/java/io/vertx/grpc/server/GrpcServerOptions.java
tsegismont: We have methods like this one in many option classes (e.g. addXXX when there is a list to manage). It's not broken since we have get/set above.

vertx-grpc-common/src/main/java/io/vertx/grpc/common/impl/GrpcMessageImpl.java
tsegismont: ok

vertx-grpc-server/src/main/java/io/vertx/grpc/server/impl/GrpcServerResponseImpl.java
tsegismont: ok

vertx-grpc-server/src/main/java/io/vertx/grpc/server/impl/GrpcServerResponseImpl.java
tsegismont: Thanks for the tip

tsegismont:
@vietj PTAL

vertx-grpc-server/src/main/java/io/vertx/grpc/server/impl/GrpcServerImpl.java
kureuil, line 75: Is there a vert.x specific reason for not allowing gRPC-Web over HTTP/2 ?
I couldn't find a reference to this restriction in the [protocol specification](https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-WEB.md#protocol-differences-vs-grpc-over-http2), and Envoy's filter doesn't seem to refuse HTTP/2 request for gRPC-Web.

vietj: what would be the use case of using gRPC web over HTTP/2 ?

tsegismont: > Is there a vert.x specific reason for not allowing gRPC-Web over HTTP/2 ?

@kureuil  Not really. I did not give enough though to this, I guess.

> what would be the use case of using gRPC web over HTTP/2 ?

@vietj gRPC in the browser over HTTP/2.

The browser does not expose protocol details, so even if the client and server communicate over HTTP/2, standard gRPC cannot be used and gRPC-Web is mandatory.

tsegismont: Thanks @kureuil for bringing this up

vietj: my question was implicitely: given that gRPC over HTTP/2 is not possible in the browser, what would be the other use case of gRPC Web over HTTP/2 that is not the browser :-)

kureuil: I only have second-hand knowledge on the matter, but one other use case for gRPC-Web over HTTP/2 would be iOS applications : the standard URLSession doesn't support HTTP trailers. Applications communicating with a gRPC server usually include the SwiftNIO package, adding more than 2Mo of compiled code and a more complex network stack "just" to support trailers in the common case. The gRPC-Web protocol allows using the standard network stack, leveraging the battery- and network-savings optimizations included in the operating system.

The use case that prompted my question is gRPC Web over HTTP/2 in the browser :)

vertx-grpc-common/src/main/java/io/vertx/grpc/common/impl/GrpcMessageImpl.java
```diff
@@ -38,11 +38,15 @@ public Buffer payload() {
   }
 
   public static Buffer encode(GrpcMessage message) {
+    return encode(message, false);
+  }
+
+  public static BufferInternal encode(GrpcMessage message, boolean trailer) {
```
vietj, line 44: add a comment mentionnning that the boolean trailer is specific to grpc web

tsegismont:
@vietj PTAL

vietj (approved):

PR #90, opened 2024-02-28
Provide common attributes in gRPC bridge ServerCallImpl

kureuil:
Despite the `ServerCall::getAttributes()` API being marked "experimental", it seems to be the established way of fetching the network addresses of the current connection. It is used in various examples found online and also in the server interceptor included in OpenTelemetry.

I didn't put the test in the `ServerTest` base class because it only concerns the GrpcBridge mode.

Conformance:

### DIFF
--- a/vertx-grpc-server/src/main/java/io/vertx/grpc/server/impl/GrpcServiceBridgeImpl.java
+++ b/vertx-grpc-server/src/main/java/io/vertx/grpc/server/impl/GrpcServiceBridgeImpl.java
@@ -10,10 +10,12 @@
  */
 package io.vertx.grpc.server.impl;
 
+import io.grpc.Attributes;
 import io.grpc.Compressor;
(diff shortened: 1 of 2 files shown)


vietj:
thanks @kureuil , does this applies to client calls as well ?

kureuil:
I tried to implement the change for the client calls, but it is not possible yet : there is no `GrpcClientBridge` that would allow one to use grpc-java `ClientInterceptor` which is the API exposing `ClientCall`.

wdyt @vietj ?

vietj:
I have no idea yet :-)

kureuil:
Having thought about this, I don't think the implementation for client calls should be in this PR. A `GrpcClientBridge` implementation would deserve its own PR, and I'd prefer to keep the scope of this one as reasonable as possible.

vietj:
can you make a backport PR to 4.x branch @kureuil ?

PR #115, opened 2024-09-04
gRPC reflection v1 support

### LINKED ISSUE #22: ProtoReflectionService "server" is null exception
### Version

4.3.5

### Context

Encountering exception when querying grpc reflection service

Testing with grpcurl:
`grpcurl -plaintext localhost:8080list`

Exception:
`
15:01:35.948 [vert.x-eventloop-thread-2] ERROR io.vertx.core.impl.ContextBase - Unhandled exception
java.lang.NullPointerException: Cannot invoke "io.grpc.Server.getImmutableServices()" because "server" is null
	at io.grpc.protobuf.services.ProtoReflectionService.getRefreshedIndex(ProtoReflectionService.java:93)
	at io.grpc.protobuf.services.ProtoReflectionService.serverReflectionInfo(ProtoReflectionService.java:134)
	at io.grpc.reflection.v1alpha.ServerReflectionGrpc$MethodHandlers.invoke(ServerReflectionGrpc.java:208)
	at io.grpc.stub.ServerCalls$StreamingServerCallHandler.startCall(ServerCalls.java:235)
	at io.vertx.grpc.server.impl.GrpcServiceBridgeImpl.lambda$bind$2(GrpcServiceBridgeImpl.java:62)
	at io.vertx.grpc.server.impl.GrpcServerImpl$MethodCallHandler.handle(GrpcServerImpl.java:95)
	at io.vertx.grpc.server.impl.GrpcServerImpl.handle(GrpcServerImpl.java:62)
`

```kotlin
val grpcServer: GrpcServer = GrpcServer.server(vertx)
val mainBridge = GrpcServiceBridge.bridge(MainBridge())
val reflectionBridge = GrpcServiceBridge.bridge(ProtoReflectionService.newInstance())
mainBridge.bind(grpcServer)
reflectionBridge.bind(grpcServer)

httpServer = vertx
  .createHttpServer(httpOptions)
  .requestHandler(grpcServer)
  //.requestHandler(router)
  .listen(appConfig.webPort).await()
```

wvuong:
Motivation:

Add support for the gRPC reflection api which is commonly used by gRPC tooling like Postman, Insomnia, k6, grpcurl for service discovery and transcoding. Fixes #22 

* Added the v1 reflection.proto from gRPC
* Updated pom.xml to compile reflection.proto
* Added GrpcServerIndex which contains service descriptors
* Added ReflectionServiceV1Handler which is a Vertx gRPC handler for handling reflection requests
* Added ReflectionServiceV1Test to test and demonstrate usage

Conformance:

Eclipse Contributor Agreement signed.

### DIFF
--- a/vertx-grpc-docs/src/main/asciidoc/ioserver.adoc
+++ b/vertx-grpc-docs/src/main/asciidoc/ioserver.adoc
@@ -38,3 +38,23 @@ The Vert.x gRPC Server can bridge a gRPC service to use with _grpc-java_ generat
 ----
 
 The bridge supports deadline automatic cancellation: when a gRPC request carrying a timeout is received, a deadline is associated with the `io.grpc.Context` an can be obtained from the current context. This deadline automatically cancels the request in progress when its associated timeout fires.
+
+=== gRPC Reflection APIs
+
+Support for the https://grpc.io/docs/guides/reflection/[gRPC reflection APIs] can be added to your Vert.x gRPC Server.
+
+[source,java]
+----
+{@link examples.GrpcIoServerExamples#reflectionExample}
+----
+
+You can then use tools like https://github.com/fullstorydev/grpcurl[gRPCurl] to explore and invoke your gRPC APIs.
+
+[source,bash]
+----
+grpcurl -plaintext localhost:50051 list
+
+grpcurl -plaintext localhost:50051 describe .helloworld.HelloRequest
+
+grpcurl -plaintext -d '{"name": "Vert.x"}' localhost:50051 helloworld.Greeter
+----
--- a/vertx-grpc-docs/src/main/java/examples/GrpcIoServerExamples.java
+++ b/vertx-grpc-docs/src/main/java/examples/GrpcIoServerExamples.java
@@ -1,5 +1,6 @@
 package examples;
 
+import io.grpc.reflection.v1.ServerReflectionGrpc;
 import io.grpc.stub.StreamObserver;
 import io.vertx.core.Vertx;
 import io.vertx.core.http.HttpServer;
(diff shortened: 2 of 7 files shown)


vietj:
can we try to make this work for vertx-grpc-server as well ?

wvuong:
Yes, we have that working on the 4.x branch at my job. However for 5.x, GrpcServerIndex is dependent on a few classes from `io.grpc.grpc-api` to parse out the service descriptors. And it looks like you are trying to keep a separation and don't want to add `io.grpc.grpc-api` as a direct dependency in vertx-grpc-server...

Perhaps this is better as a new submodule in the project so that people can use it with vertx-grpc-server or vertx-grpcio-server?

vietj:
which API is necessary ? vertx has its own generator so perhaps we can have this working as well using the generator.

vietj:
I think I will merge it as is and take it from here as I need it with pure gRPC to build JPMS/gRPC example

wvuong:
for GrpcServerIndex
```
import io.grpc.ServerServiceDefinition;
import io.grpc.ServiceDescriptor;
```
from `io.grpc:grpc-api`

vietj:
I think actually for now what we are missing is documentation, can you add this ? it is very important

wvuong:
I can update ioserver.adoc and add an example tonight

vietj:
thanks @wvuong

zeval:
Hey everyone. What's missing for this to go forward? Reflection would be very welcome for our use-case with this library.

vietj:
I think we would need a rebase on the latest master commit

vietj:
I think we should also try to make it work with the grpc server instead of the grpcio server but that is an implementation detail

PR #120, opened 2024-12-19
case insensitive grpc headers/metadata

hcura:
When using grpc web (with HTTP 1.1), the http headers are not lower cased when converting into grpc metadata. This means that they become inaccessible inside metadata since all the keys are converted to lower case on creation.

This erratic behaviour was observed in a couple of browsers that did not respect a lower case `authorization` header and transmitted upper case over the wire (confirmed on Wireshark).

This works fine for http2 since it's supposed to provide stronger case insensitive guarantees.

The test I added fails, if the `.toLowerCase()` is removed.

### DIFF
--- a/vertx-grpcio-common/src/main/java/io/vertx/grpcio/common/impl/Utils.java
+++ b/vertx-grpcio-common/src/main/java/io/vertx/grpcio/common/impl/Utils.java
@@ -57,7 +57,7 @@ public class Utils {
     byte[][] array = new byte[entries.size() * 2][];
     int idx = 0;
     for (Map.Entry<String, String> entry : entries) {
-      String key = entry.getKey();
+      String key = entry.getKey().toLowerCase();
       array[idx++] = key.getBytes(StandardCharsets.UTF_8);
       String value = entry.getValue();
       byte[] data;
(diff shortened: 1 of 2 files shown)


vertx-grpcio-common/src/main/java/io/vertx/grpcio/common/impl/Utils.java
```diff
@@ -57,7 +57,7 @@ public static Metadata readMetadata(MultiMap headers) {
     byte[][] array = new byte[entries.size() * 2][];
     int idx = 0;
     for (Map.Entry<String, String> entry : entries) {
-      String key = entry.getKey();
+      String key = entry.getKey().toLowerCase();
```
tsegismont, line 60: `toLowerCase()` is locale sensitive, this should be:

```java
      String key = entry.getKey().toLowerCase(Locale.ROOT);
```

tsegismont (changes requested):
Thank you for trying out Vert.x gRPC Web support and reporting this @hcura 

I have only one concern but overall it looks good to me.

vertx-grpcio-common/src/main/java/io/vertx/grpcio/common/impl/Utils.java
hcura: makes sense. fixed!

tsegismont (approved):
LGTM, thank you!

Issue #95, opened 2024-05-10
Grpc transcoding


zZHorizonZz:
#### Describe the feature

This feature request proposes gRPC transcoding capabilities for the server. This would allow the server to accept json formatted grpc requests and transcode them according to the specifications defined in google.api.http.

#### Use cases

While grpc web is already implemented, it utilizes a unique format that differs from the transcoding of REST requests into grpc requests. Implementing this feature would be particularly useful when publicly exposing a grpc service, where supporting web based requests or providing users with a REST api option is desired.

#### Contribution

I am willing to contribute to the implementation of this feature. I have already implemented this in quarkus repository. After discussion in [this PR](https://github.com/quarkusio/quarkus/pull/40254), the maintainers concluded that this should be implemented directly in grpc server implementation.

vietj:
@zZHorizonZz we would be pleased to have this feature, perhaps we can discuss how the implementation should be crafted for the vertx-grpc project ?

vietj:
we are planning to backport grpc-web to vertx 4, so we can assume developing your PR against master can be done and then we can backport everything to vertx 4.

zZHorizonZz:
I will need to go through the code of grpc-web so that I understand it better, as I reckon the transcoding implementation will be using some elements from grpc-web.

vietj:
sure @zZHorizonZz , there might be some decoupling / refactoring necessary because the current grpc-web integration uses inheritance with calls to super to override the behaviour for grpc-web sometimes

vietj:
we can also schedule a call to discuss the plans if you are inclined

zZHorizonZz:
I wouldn't mind getting into a call, although I will be busy until the end of the month, so it would be in 2-3 weeks.

zZHorizonZz:
I think I should have more time this year, as last year I didn't have as much free time as I would have liked to work on this. I still think it will take some time (even more because I need to remember how the actual thing I wrote half a year ago works... 😅)

Anyway, my initial thought process on how this could be implemented is that since gRPC transcoding supports annotations like this:

```
option (google.api.http) = {
    post: "/v1/simple"
    body: "*"
};
```

I think we should start implementing support for this in the generator, if it isn't already supported or if there aren't other protoc plugins available.

PR #122, opened 2025-01-22
Add support for gRPC transcoding

zZHorizonZz:
Motivation:

This pull request introduces gRPC transcoding capabilities to the gRPC server. The most important changes include the addition of new interfaces for HTTP templates and variable bindings, as well as updates to existing classes to support these new features.

The code for path lookup and checking is based on code from [grpc-transcoding-repo](https://github.com/grpc-ecosystem/grpc-httpjson-transcoding). I think, we shouldn't try to reinvent the wheel when a solution already exists. While there are some changes since the original code is in C++, the logic for matching and extracting HTTP requests remains largely similar. I have also converted several tests from the original repository for these classes.

Regarding the actual transcoding, ~~I'm currently facing two issues. First, by definition, requests can have an empty body while the input gRPC message can still have fields. What I mean is that values can be passed in requests via path or query parameters. However, when I test this, the code gets stuck, my guess is that it's possibly waiting for a body. I need toinvestigate this further but would appreciate any help. The second issue is that in the generator, the code seems unable to read method options despite having them as dependencies~~. But, overall the transcoding seems to be working except for these two issues, though obviously, this pr will need additional work.

Also currently there isn't really support for streaming that would require websockets which is an entirely new thing in my opinion.

The code for the transcoding is reusing the wire format infrastructure by putting length and compression data before the json data, so that we don't need to create entirely new thing for this.

I appreciate any feedback and help with integrating this feature into the codebase.

Closes: https://github.com/eclipse-vertx/vertx-grpc/issues/95


vietj:
I need to read more about it, however I think this feature should be implemented in its own module instead of being in grpc common

vietj:
I am wondering if we should not make some work on the server to provide an invocation API to decouple the delivery of gRPC invocations from the actual transport.

zZHorizonZz:
Are you talking about implementing a middleware layer where you could have daisy chained middlewares with buffers or something, so you could have compression middleware, transcoding middleware, etc.? Or am I thinking of something completely different?

vietj:
I'm thinking of a way to community between a gRPC server request/response and an invocation layer using grpc messages, this way anything could invoke a gRPC service by crafting the proper messages, very much like HTTP messages in Netty

zZHorizonZz:
I think I might be understanding it, but just maybe....

So if I understand it basically there would be something like this:
```java
public interface MethodRegistry {

<Req, Resp> MethodRegistry callHandler(ServiceMethod<Req, Resp> serviceMethod, Handler<GrpcRequest<Req, Resp>> handler);

}
```

```java
public interface MethodCallHandler<Req, Resp> implements Handler<GrpcRequest<Req, Resp>> {

void handle(GrpcRequest<Req, Resp> grpcRequest);

}
```

So here default implementation if method registry would contain basically logic from server:
```java
if (handler != null) {
      MethodCallHandler<Req, Resp> p = new MethodCallHandler<>(serviceMethod.decoder(), serviceMethod.encoder(), handler);
      methodCallHandlers.compute(serviceMethod.fullMethodName(), (key, prev) -> {
        if (prev == null) {
          prev = new ArrayList<>();
        }
        for (int i = 0; i < prev.size(); i++) {
          MethodCallHandler<?, ?> a = prev.get(i);
          if (a.messageDecoder.format() == serviceMethod.decoder().format() && a.messageEncoder.format() == serviceMethod.encoder().format()) {
            prev.set(i, p);
            return prev;
          }
        }
        prev.add(p);
        return prev;
      });
    } else {
      methodCallHandlers.compute(serviceMethod.fullMethodName(), (key, prev) -> {
        if (prev != null) {
          for (int i = 0; i < prev.size(); i++) {
            MethodCallHandler<?, ?> a = prev.get(i);
            if (a.messageDecoder.format() == serviceMethod.decoder().format() && a.messageEncoder.format() == serviceMethod.encoder().format()) {
              prev.remove(i);
              break;
            }
          }
          if (prev.isEmpty()) {
            prev = null;
          }
        }
        return prev;
      });
    }
    return this;

File: vertx-grpc-server/src/main/java/io/vertx/grpc/server/impl/GrpcServerImpl.java [hunk 1]
```diff
@@ -206,6 +282,24 @@ public <Req, Resp> GrpcServer callHandler(ServiceMethod<Req, Resp> serviceMethod
     return this;
   }
 
+  @Override
+  public <Req, Resp> GrpcServer callHandlerWithTranscoding(ServiceMethod<Req, Resp> serviceMethod, Handler<GrpcServerRequest<Req, Resp>> handler,
+    ServiceTranscodingOptions transcodingOptions) {
+    this.callHandler(serviceMethod, handler);
+
+    if (!options.isGrpcTranscodingEnabled()) {
```
vietj, line 290: this should throw IllegalStateException

vietj:
I'll handle the internal refactoring (decoupling) myself after we merge this PR

vertx-grpc-transcoding/src/main/java/io/vertx/grpc/transcoding/HttpTemplate.java
```diff
@@ -0,0 +1,59 @@
...
+
+    return new HttpTemplateImpl(parser.segments(), parser.verb(), parser.variables());
+  }
+
+  /**
+   * Returns the list of segments in the parsed template.
+   *
+   * @return The list of segments.
+   */
+  List<String> getSegments();
```
vietj, line 44: rename `segments`

vertx-grpc-transcoding/src/main/java/io/vertx/grpc/transcoding/HttpTemplate.java
```diff
@@ -0,0 +1,59 @@
...
+
+  /**
+   * Returns the verb in the parsed template.
+   *
+   * @return The verb.
+   */
+  String getVerb();
```
vietj, line 51: rename `verb`

vertx-grpc-transcoding/src/main/java/io/vertx/grpc/transcoding/HttpTemplate.java
```diff
@@ -0,0 +1,59 @@
...
+
+  /**
+   * Returns the list of variables in the parsed template.
+   *
+   * @return The list of variables.
+   */
+  List<HttpTemplateVariable> getVariables();
```
vietj, line 58: rename `variables`

vertx-grpc-transcoding/src/main/java/io/vertx/grpc/transcoding/HttpTemplateVariable.java
```diff
@@ -0,0 +1,91 @@
...
+    variable.setWildcardPath(wildcardPath);
+    return variable;
+  }
+
+  /**
+   * Returns the field path of the variable.
+   *
+   * @return The field path.
+   */
+  List<String> getFieldPath();
```
vietj, line 41: rename `fieldPath`

vertx-grpc-transcoding/src/main/java/io/vertx/grpc/transcoding/HttpTemplateVariable.java
```diff
@@ -0,0 +1,91 @@
...
+   * @param fieldPath The field path to set.
+   */
+  void setFieldPath(List<String> fieldPath);
+
+  /**
+   * Returns the starting segment of the variable.
+   *
+   * @return The starting segment.
+   */
+  int getStartSegment();
```
vietj, line 55: rename `startSegment`

vertx-grpc-transcoding/src/main/java/io/vertx/grpc/transcoding/HttpTemplateVariable.java
```diff
@@ -0,0 +1,91 @@
...
+   * @param startSegment The starting segment to set.
+   */
+  void setStartSegment(int startSegment);
+
+  /**
+   * Returns the ending segment of the variable.
+   *
+   * @return The ending segment.
+   */
+  int getEndSegment();
```
vietj, line 69: rename `endSegment`

vertx-grpc-transcoding/src/main/java/io/vertx/grpc/transcoding/HttpTemplateVariable.java
```diff
@@ -0,0 +1,91 @@
...
+   * @param endSegment The ending segment to set.
+   */
+  void setEndSegment(int endSegment);
+
+  /**
+   * Checks if the variable represents a wildcard path.
+   *
+   * @return {@code true} if the variable represents a wildcard path, {@code false} otherwise.
+   */
+  boolean hasWildcardPath();
```
vietj, line 83: rename `isWildCardPath`

vertx-grpc-transcoding/src/main/java/io/vertx/grpc/transcoding/HttpVariableBinding.java
```diff
@@ -0,0 +1,60 @@
...
+  static HttpVariableBinding create(List<String> fieldPath, String value) {
+    return new HttpVariableBindingImpl(fieldPath, value);
+  }
+
+  /**
+   * Gets the field path that describes where in the gRPC message the value should be placed.
+   *
+   * @return A list of field names representing the path in the gRPC message
+   */
+  List<String> getFieldPath();
```
vietj, line 38: rename `fieldPath`

vertx-grpc-transcoding/src/main/java/io/vertx/grpc/transcoding/HttpVariableBinding.java
```diff
@@ -0,0 +1,60 @@
...
+   * @param fieldPath A list of field names representing the path in the gRPC message
+   */
+  void setFieldPath(List<String> fieldPath);
+
+  /**
+   * Gets the value that was extracted from the HTTP request.
+   *
+   * @return The string value to be bound to the specified field path
+   */
+  String getValue();
```
vietj, line 52: rename `value`

vertx-grpc-transcoding/src/main/java/io/vertx/grpc/transcoding/PathMatcherBuilder.java
```diff
@@ -0,0 +1,118 @@
...
+   * @return true if registration was successful, false otherwise
+   */
+  boolean register(ServiceTranscodingOptions transcodingOptions, String method);
+
+  /**
+   * Gets the root node of the path matching tree.
+   *
+   * @return the root PathMatcherNode
+   */
+  PathMatcherNode getRoot();
```
vietj, line 40: rename `root`

vertx-grpc-transcoding/src/main/java/io/vertx/grpc/transcoding/PathMatcherBuilder.java
```diff
@@ -0,0 +1,118 @@
...
+
+  /**
+   * Gets the set of registered custom verbs.
+   *
+   * @return set of custom verb strings
+   */
+  Set<String> getCustomVerbs();
```
vietj, line 47: rename `customVerbs`

vertx-grpc-transcoding/src/main/java/io/vertx/grpc/transcoding/PathMatcherBuilder.java
```diff
@@ -0,0 +1,118 @@
...
+
+  /**
+   * Gets the list of registered method data.
+   *
+   * @return list of PathMatcherMethodData objects
+   */
+  List<PathMatcherMethodData> getMethodData();
```
vietj, line 54: rename `methodData`

vertx-grpc-transcoding/src/main/java/io/vertx/grpc/transcoding/PathMatcherBuilder.java
```diff
@@ -0,0 +1,118 @@
...
+   * @param pathUnescapeSpec the URL unescaping specification to use
+   */
+  void setUrlUnescapeSpec(PercentEncoding.UrlUnescapeSpec pathUnescapeSpec);
+
+  /**
+   * Gets the current URL unescaping specification.
+   *
+   * @return the current URL unescaping specification
+   */
+  PercentEncoding.UrlUnescapeSpec getUrlUnescapeSpec();
```
vietj, line 68: rename `urlUnescapeSpec`

vertx-grpc-transcoding/src/main/java/io/vertx/grpc/transcoding/PathMatcherBuilder.java
```diff
@@ -0,0 +1,118 @@
...
+   * @param queryParamUnescapePlus true to unescape plus signs, false otherwise
+   */
+  void setQueryParamUnescapePlus(boolean queryParamUnescapePlus);
+
+  /**
+   * Gets whether plus signs in query parameters are being unescaped to spaces.
+   *
+   * @return current setting for plus sign unescaping in query parameters
+   */
+  boolean getQueryParamUnescapePlus();
```
vietj, line 82: rename `queryParamUnescapePlus`

vertx-grpc-transcoding/src/main/java/io/vertx/grpc/transcoding/PathMatcherBuilder.java
```diff
@@ -0,0 +1,118 @@
...
+   * @param matchUnregisteredCustomVerb true to match unregistered custom verbs, false otherwise
+   */
+  void setMatchUnregisteredCustomVerb(boolean matchUnregisteredCustomVerb);
+
+  /**
+   * Gets whether unregistered custom verbs are being matched.
+   *
+   * @return current setting for matching unregistered custom verbs
+   */
+  boolean getMatchUnregisteredCustomVerb();
```
vietj, line 96: rename `matchUnregisteredCustomVerb`

vertx-grpc-transcoding/src/main/java/io/vertx/grpc/transcoding/PathMatcherBuilder.java
```diff
@@ -0,0 +1,118 @@
...
+   * @param failRegistrationOnDuplicate true to fail on duplicates, false to silently continue
+   */
+  void setFailRegistrationOnDuplicate(boolean failRegistrationOnDuplicate);
+
+  /**
+   * Gets whether registration fails on duplicate patterns.
+   *
+   * @return current setting for failing on duplicate registrations
+   */
+  boolean getFailRegistrationOnDuplicate();
```
vietj, line 110: rename `failRegisrationOnDuplicate`

vertx-grpc-transcoding/src/main/java/io/vertx/grpc/transcoding/PathMatcherLookupResult.java
```diff
@@ -0,0 +1,27 @@
...
+  private final List<HttpVariableBinding> variableBindings;
+  private final String bodyFieldPath;
+
+  public PathMatcherLookupResult(String method, List<HttpVariableBinding> variableBindings, String bodyFieldPath) {
+    this.method = method;
+    this.variableBindings = variableBindings;
+    this.bodyFieldPath = bodyFieldPath;
+  }
+
+  public String getMethod() {
```
vietj, line 16: rename `method`

vertx-grpc-transcoding/src/main/java/io/vertx/grpc/transcoding/PathMatcherLookupResult.java
```diff
@@ -0,0 +1,27 @@
...
+    return method;
+  }
+
+  public List<HttpVariableBinding> getVariableBindings() {
```
vietj, line 20: rename `variableBindings`

vertx-grpc-transcoding/src/main/java/io/vertx/grpc/transcoding/PathMatcherLookupResult.java
```diff
@@ -0,0 +1,27 @@
...
+    return variableBindings;
+  }
+
+  public String getBodyFieldPath() {
```
vietj, line 24: rename `bodyFieldPath`

vertx-grpc-transcoding/src/main/java/io/vertx/grpc/transcoding/ServiceTranscodingOptions.java
```diff
@@ -0,0 +1,132 @@
...
+ * including path templates, HTTP methods, body mappings, and additional bindings.
+ *
+ * The transcoding options define how incoming HTTP requests should be mapped to gRPC method calls, including:
+ * <ul>
+ *   <li>Which gRPC service/method to call (selector)</li>
+ *   <li>What HTTP method and path pattern to match</li>
+ *   <li>How to map the HTTP request/response bodies to gRPC messages</li>
+ * </ul>
+ */
+@VertxGen
```
vietj, line 19: This should be `@DataObject` instead in addition of its components.

We should also make those interfaces plain classes as they are rather data containers and not asynchronous and annotation all of them with `@DataObject`

vietj:
Most objects in the transcoding package should be annotated with `@DataObject` and should be plain classes instead of having an interface/implementation

zZHorizonZz:
> I'll handle the internal refactoring (decoupling) myself after we merge this PR

I wish I could help, but I think I'm imagining something completely different from what you're thinking, so that will probably be best.

> Most objects in the transcoding package should be annotated with @DataObject and should be plain classes instead of having an interface/implementation

Yeah, I have refactored it so now there are just classes. I did it mostly to fit common package style and also because I had some issues compiling, as I was getting errors about @VertxGen usage, etc., as I've never really used these annotations. But now since it's in its own package, there are no issues.

vertx-grpc-docs/src/main/asciidoc/server.adoc
```diff
@@ -68,6 +68,12 @@ router.route("/com.mycompany.MyService/*").handler(corsHandler);
 ----
 ====
 
+==== gRPC Transcoding
+
+The Vert.x gRPC Server supports the gRPC transcoding. The transcoding is disabled by default.
```
vietj, line 73: link to grpc transcoding

zZHorizonZz: Most of the time when I need to read about transcoding I'm going to google cloud grpc docs as I didn't really find any official transcoding docs. So is this [link](https://cloud.google.com/endpoints/docs/grpc/transcoding) ok are are you have something else in mind.

vietj: My point is that the documentation should do more and explain what transcoding does and when it is useful for users.

zZHorizonZz:
I have added documentation. While writing docs for the generator, I have been thinking and this is more or less for discussion if we should add configuration options to enable/disable the generation of wire format and transcoding options. I think in large codebases where these features are not needed, it might make a significant difference in terms of JAR file size.

vertx-grpc-server/src/main/java/io/vertx/grpc/server/impl/GrpcServerImpl.java
```diff
@@ -106,37 +133,82 @@ public void handle(HttpServerRequest httpRequest) {
...
+
+    if (GrpcProtocol.HTTP_1.mediaType().equals(request.headers().get(CONTENT_TYPE)) && pathMatcherLookupResult != null) {
+      return new GrpcMethodCall("/" + pathMatcherLookupResult.getMethod());
+    }
+
+    return new GrpcMethodCall(request.path());
+  }
+
+  private <Req, Resp> void handle(MethodCallHandler<Req, Resp> method, HttpServerRequest request, GrpcMethodCall methodCall, PathMatcherLookupResult pathMatcherLookupResult) {
+    if (request.version() == HttpVersion.HTTP_2) {
```
vietj, line 177: why do we need to not accept HTTP/2 requests?

vertx-grpc-server/src/main/java/io/vertx/grpc/server/impl/GrpcServerImpl.java
```diff
@@ -106,37 +133,82 @@ public void handle(HttpServerRequest httpRequest) {
...
+      return;
+    }
+
+    String contentType = request.getHeader(CONTENT_TYPE);
```
vietj, line 181: we should send an appropriate HTTP response

vertx-grpc-server/src/main/java/io/vertx/grpc/server/impl/GrpcServerImpl.java
```diff
@@ -106,37 +133,82 @@ public void handle(HttpServerRequest httpRequest) {
...
+      return;
+    }
+
+    List<HttpVariableBinding> bindings = new ArrayList<>();
+    if (request.version() != HttpVersion.HTTP_2 && GrpcProtocol.HTTP_1.mediaType().equals(request.getHeader(CONTENT_TYPE))) {
+      bindings.addAll(pathMatcherLookupResult.getVariableBindings());
+    }
+
+    MethodTranscodingOptions transcodingOptions = this.transcodingOptions.get(methodCall.fullMethodName());
+    if (transcodingOptions == null) {
```
vietj, line 192: we should send an appropriate HTTP response

vertx-grpc-server/src/main/java/io/vertx/grpc/server/impl/GrpcServerRequestImpl.java
```diff
@@ -213,6 +243,61 @@ private boolean notGrpcWebText() {
     return grpcWebTextBuffer == null;
   }
 
+  public boolean isTranscodable() {
+    return (httpRequest.version() == HttpVersion.HTTP_1_0 || httpRequest.version() == HttpVersion.HTTP_1_1) && GrpcProtocol.HTTP_1.mediaType()
+      .equals(httpRequest.getHeader(CONTENT_TYPE));
+  }
+
+  private Buffer mutateMessage(Buffer message, List<HttpVariableBinding> bindings) {
```
vietj, line 251: we should move this code to the transcoding module so it can be tested there independently  of gRPC interactions.

vertx-grpc-server/src/main/java/io/vertx/grpc/server/impl/GrpcServerImpl.java
zZHorizonZz: We don't, I will remove that 👍 I think I initially had this just as check so somebody by accident cannot call this in the future.

vietj:
> I have added documentation. While writing docs for the generator, I have been thinking and this is more or less for discussion if we should add configuration options to enable/disable the generation of wire format and transcoding options. I think in large codebases where these features are not needed, it might make a significant difference in terms of final file size.

we should only perform transcoding generation when the service is annotated with `google.api.http` so the intent is clear

zZHorizonZz:
> we should only perform transcoding generation when the service is annotated with google.api.http so the intent is clear

So basically, currently the generator is generating MethodTranscodingOptions for every method and is using default values in these. For example, for the HTTP method it falls back to POST, and for the path it falls back to a path like this:

```java
if (methodContext.transcodingContext.path == null) {
        methodContext.transcodingContext.method = "POST";
        if (serviceContext.packageName == null || serviceContext.packageName.isEmpty()) {
          methodContext.transcodingContext.path = "/" + serviceContext.serviceName + "/" + methodContext.methodName;
        } else {
          methodContext.transcodingContext.path = "/" + serviceContext.packageName + "/" + serviceContext.serviceName + "/" + methodContext.methodName;
        }
      }
```

I think I have an idea on how to generate the code only when there are transcoding annotations available though, so I will try to fix that.

> is grpc.io server generation capable of doing transcoding ? I don't see any the service options metadata available in the generated code

If we are talking about server from for example vertx-grpc-it then there should be generated code on top of the class something like this:
```java
public static final MethodTranscodingOptions SayHello_TRANSCODING = new MethodTranscodingOptions(
    "",
    HttpMethod.valueOf("POST"),
    "/Greeter/SayHello",
    "",
    "",
    List.of(
    ));
 ```
 
 And then some binding methods for transcoding.

**Comment by vietj** [comment]
about grpc.io, I am merely wondering whether the transcoding information is available from the grpc-java generator, because I cannot see anything here, so there is no possibility to take a grpc-java service descriptor and obtain the `oogle.api.http` information for honnoring the declared mapping.

**Comment by vietj** [comment]
for the vertx generator, we need to honour the `google.api.http` to implement it, that is what users will expect to use

**Comment by zZHorizonZz** [comment]
By grpc.io do you mean vertx-grpcio-server module by any chance? I didn't really do anything with that.

**Comment by vietj** [comment]
yes I mean vertx-grpcio-server (which uses grpc-java generated code), I am
not asking for anything do be done here (yet) because it seems to me that
the grpc.io generated code does not provide for such options

**Comment by zZHorizonZz** [comment]
Ok, so I have fixed the second issue, I had with the generator, so now the generator is generating the transcoding options only for annotated methods. The issue was that I forgot you need to pass the list of options it should be able to read.

**Comment by zZHorizonZz** [comment]
Also, I don't know if I should handle this on my own or if you can do that after this is merged, but the files currently don't have license headers, I have included the source of the original files from the gRPC transcoding C++ repo (basically HTTP template classes and path matcher are based on the code from that)  that I rewrote in Java. I'm not sure whether the Apache license of the original repo needs to be defined in the file headers alongside the Eclipse license.

**Comment by vietj** [comment]
@zZHorizonZz can you do a clean rebase on latest repo main branch ? that makes it easier to merge and review

**Comment by vietj** [comment]
it seems it is doing good progress too :-)

File: vertx-grpc-it/src/test/proto/helloworld.proto [hunk 27]
```diff
@@ -39,7 +41,9 @@ package helloworld;
 // The greeting service definition.
 service Greeter {
   // Sends a greeting
-  rpc SayHello (HelloRequest) returns (HelloReply) {}
+  rpc SayHello (HelloRequest) returns (HelloReply) {
+    option (google.api.http) = { get: "/v1/hello/{name}" };
```
vietj, line 45: looks great, we need tests for other cases

zZHorizonZz:
> @zZHorizonZz can you do a clean rebase on latest repo main branch ? that makes it easier to merge and review

Should I also squash the commits? I don't know what the view on that is in Eclipse projects, but in Quarkus they want that to be done so the git history is more or less clean.

vietj:
> > @zZHorizonZz can you do a clean rebase on latest repo main branch ? that makes it easier to merge and review
> 
> Should I also squash the commits? I don't know what the view on that is in Eclipse projects, but in Quarkus they want that to be done so the git history is more or less clean.

we don't mind having commits not squashed if they are properly separated, there is value in the history we might want to preserve sometimes.

we can check when we merge the contribution

vietj:
your commit history looks clean, so we might preserve it

vietj:
I can see many code reformatted by your IDE which makes the PR harder to review, can you remove those un-necessary changes ?

zZHorizonZz:
> I can see many code reformatted by your IDE which makes the PR harder to review, can you remove those un-necessary changes ?

Yeah, sorry about that. IDE is doing its thing where it seems like it's able to adapt to the code style, but at the same time, it's not. I'll fix that.

vietj:
@zZHorizonZz I'll make another review soonish, hopefully we can get that in time for 5.0

vertx-grpc-transcoding/src/main/java/module-info.java
```diff
@@ -0,0 +1,13 @@
+module io.vertx.grpc.transcoding {
+  requires io.netty.common;
+  requires io.netty.buffer;
+  requires io.netty.codec;
+  requires io.netty.codec.compression;
+  requires io.netty.transport;
+  requires com.google.protobuf;
+  requires com.google.protobuf.util;
+  requires transitive io.vertx.core;
```
vietj, line 9: avoid _transitive_ here, for the moment we avoid transitive dependency when declaring module descriptors

vertx-grpc-transcoding/src/main/java/io/vertx/grpc/transcoding/PercentEncoding.java
```diff
@@ -0,0 +1,133 @@
...
+
+    return unescaped.toString();
+  }
+
+  public static String urlUnescapeString(String part) {
+    return urlUnescapeString(part, UrlUnescapeSpec.ALL_CHARACTERS, false);
+  }
+
+  public static String urlEscapeString(String str) {
+    return Pattern.compile("[^a-zA-Z0-9-_.~]").matcher(str)
```
vietj, line 130: Cache pattern instance in a static final declaration.

vietj:
do we have a test for sub query parameters like here https://github.com/googleapis/googleapis/blob/master/google/api/http.proto#L109 ?

vertx-grpc-transcoding/src/main/java/io/vertx/grpc/transcoding/PercentEncoding.java
zZHorizonZz: I think we don't need this method at all. It's just that it was in the original file, but I think we might keep it if there's a need for it in the near future. If not, we can remove it later. But I will fix the declaration.

zZHorizonZz:
> do we have a test for sub query parameters like here https://github.com/googleapis/googleapis/blob/master/google/api/http.proto#L109 ?

There is a test in MessageWeaverTest, but not in the actual request testing. I have added a test for that and fixed an issue I found along the way.

vietj:
thanks for the contribution I plan to merge it soon

zZHorizonZz:
I remembered that I never implemented the HTTP error code mapping, so I have added that. Since there isn't much documentation, it's done on a best effort basis. I followed the mapping from the [.NET implementation of transcoding.](https://github.com/dotnet/aspnetcore/blob/81b289460a9ce5fcbd0d32d80a645f139dffc422/src/Grpc/JsonTranscoding/src/Microsoft.AspNetCore.Grpc.JsonTranscoding/Internal/JsonRequestHelpers.cs#L122)

PR #126, opened 2025-02-12
Server refactor decoupling

vietj:


vietj:
it is not clear also what "selector" means as it is currently unused

zZHorizonZz:
@vietj selector should be used to select the method to which the MethodTranscodingOptions should bind. Currently, it is unused since the current way of registering transcoding options is tied to method reference. (I will try to update this in the future to allow using google.api.http options at the service level.)

zZHorizonZz:
> @zZHorizonZz I cannot see a test for "response body", can you add such test with a PR to this branch ?

Done in https://github.com/eclipse-vertx/vertx-grpc/pull/128

PR #128, opened 2025-02-13
Grpc server refactor

zZHorizonZz:
Added unit test for response body and fixed related issue with improper content-length as requested in https://github.com/eclipse-vertx/vertx-grpc/pull/126#issuecomment-2656776534.

### DIFF
--- a/vertx-grpc-it/src/test/java/io/vertx/grpc/it/TranscodingTest.java
+++ b/vertx-grpc-it/src/test/java/io/vertx/grpc/it/TranscodingTest.java
@@ -265,6 +265,80 @@ public class TranscodingTest extends ProxyTestBase {
     test.awaitSuccess();
   }
 
+  @Test
+  public void testUnaryWithRequestBody(TestContext should) {
+    HttpClient client = vertx.createHttpClient();
+    GrpcServerOptions serverOptions = new GrpcServerOptions().setGrpcTranscodingEnabled(true);
+
+    Future<HttpServer> server = vertx.createHttpServer()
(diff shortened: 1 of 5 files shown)


vietj:
thank you can we also get an IT test with the proto file to show how it works in action with a proto file ?

vietj:
thanks, it actually is not clear what response_body stands for

zZHorizonZz:
Ohhh, I just pushed commit where I explained the whole transcoding options more in documentation, do you want me to open PR or could you put it into your branch yourself @vietj? https://github.com/zZHorizonZz/vertx-grpc/blob/grpc-server-refactor/vertx-grpc-docs/src/main/asciidoc/plugin.adoc#http-transcoding-options

vietj:
@zZHorizonZz actually I ran the test and figured out :-)

however I realised that we are missing tests that check the incorrect behaviour, e.g. using an invalid response_body that references a field that does not exist, currently this will be thrown by `weaveResponseMessage` as `IllegalStateException`.

We need to cover this with tests and perhaps change the behaviour, i.e. throw an `IllegalArgumentException` instead and most likely define/check the HTTP response behaviour.

zZHorizonZz:
Ok then I can include that in the other PR with the added tests for IllegalArgumentException :+1:

vietj:
which PR ?

besides there are other points I'd like to discuss with you about, perhaps you could join our discord server to have a conversation there ?

zZHorizonZz:
> which PR ?

I will open new PR after we discuss the details.

> besides there are other points I'd like to discuss with you about, perhaps you could join our discord server to have a conversation there ?

Yeah no problem, just joined the discord. I could not do it in the morning because I got an invalid invite but now it seems to work. :+1:

PR #136, opened 2025-03-05
Avoid un-necessary byte copies when decoding a Vert.x Buffer to a protobuf object and perform more optimal allocation

vietj:
Motivation:

The current implementation of the bridge uses a plain `ByteArrayInputStream` and then let protobuf parse the message. It turns out there is a `KnownLength` interface used by the parser to know the actual buffer size to be decoded and optimize the decoding process.

In addition we can use Netty `ByteBufInputStream` to avoid using `ByteArrayInputStream` and un-necessary copies.

Changes:

Swap usage of `ByteArrayInputStream` with a class that extends `ByteBufInputStream` and implement KnownLength.

### DIFF
--- a/vertx-grpc-common/src/main/java/io/vertx/grpc/common/impl/BridgeMessageDecoder.java
+++ b/vertx-grpc-common/src/main/java/io/vertx/grpc/common/impl/BridgeMessageDecoder.java
@@ -11,7 +11,11 @@
 package io.vertx.grpc.common.impl;
 
 import io.grpc.Decompressor;
+import io.grpc.KnownLength;
 import io.grpc.MethodDescriptor;
+import io.netty.buffer.ByteBufInputStream;
+import io.vertx.core.buffer.Buffer;
+import io.vertx.core.buffer.impl.BufferImpl;
 import io.vertx.grpc.common.GrpcMessage;
 import io.vertx.grpc.common.GrpcMessageDecoder;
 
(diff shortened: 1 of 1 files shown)


vietj:
@franz1981 ping

vertx-grpc-common/src/main/java/io/vertx/grpc/common/impl/BridgeMessageDecoder.java
```diff
@@ -29,15 +33,31 @@ public BridgeMessageDecoder(MethodDescriptor.Marshaller<T> marshaller, Decompres
     this.decompressor = decompressor;
   }
 
+  private static class KnownLengthStream extends ByteBufInputStream implements KnownLength {
+    public KnownLengthStream(Buffer buffer) {
+      super(((BufferImpl)buffer).getByteBuf(), buffer.length());
```
franz1981, line 38: the buffer is doesn't have any weird requirement re release ref cnt?

vietj: it does create a slice but closing the ByteBufInputStream releases it

vietj:
> ```
> ((BufferImpl)buffer).getByteBuf()
> ```
> 
> Casting this b/c of the method going away in Buffer interface in Vert.x5 ?

this method is available on `BufferInternal`

PR #137, opened 2025-03-17
Grpc reflection

zZHorizonZz:
This is rebase and continuation of original PR: https://github.com/eclipse-vertx/vertx-grpc/pull/115

Original PR description:
```
Motivation:

Explain here the context, and why you're making that change, what is the problem you're trying to solve.

Conformance:

```

### DIFF
--- a/vertx-grpc-server/src/main/java/module-info.java
+++ b/vertx-grpc-server/src/main/java/module-info.java
@@ -6,6 +6,7 @@ module io.vertx.grpc.server {
   requires io.vertx.codegen.api;
   requires io.netty.codec;
   requires io.netty.buffer;
+  requires com.google.protobuf;
   exports io.vertx.grpc.server;
   exports io.vertx.grpc.server.impl to io.vertx.grpc.transcoding;
 }
--- a/vertx-grpc-docs/src/main/asciidoc/ioserver.adoc
+++ b/vertx-grpc-docs/src/main/asciidoc/ioserver.adoc
@@ -38,3 +38,23 @@ The Vert.x gRPC Server can bridge a gRPC service to use with _grpc-java_ generat
 ----
 
 The bridge supports deadline automatic cancellation: when a gRPC request carrying a timeout is received, a deadline is associated with the `io.grpc.Context` an can be obtained from the current context. This deadline automatically cancels the request in progress when its associated timeout fires.
+
+=== gRPC Reflection APIs
+
+Support for the https://grpc.io/docs/guides/reflection/[gRPC reflection APIs] can be added to your Vert.x gRPC Server.
+
+[source,java]
+----
+{@link examples.GrpcIoServerExamples#reflectionExample}
+----
+
+You can then use tools like https://github.com/fullstorydev/grpcurl[gRPCurl] to explore and invoke your gRPC APIs.
+
+[source,bash]
+----
+grpcurl -plaintext localhost:50051 list
+
+grpcurl -plaintext localhost:50051 describe .helloworld.HelloRequest
+
+grpcurl -plaintext -d '{"name": "Vert.x"}' localhost:50051 helloworld.Greeter
+----
--- a/vertx-grpc-docs/src/main/java/examples/GrpcIoServerExamples.java
+++ b/vertx-grpc-docs/src/main/java/examples/GrpcIoServerExamples.java
@@ -1,5 +1,6 @@
 package examples;
 
+import io.grpc.reflection.v1.ServerReflectionGrpc;
 import io.grpc.stub.StreamObserver;
 import io.vertx.core.Vertx;
 import io.vertx.core.http.HttpServer;
@@ -7,6 +8,10 @@ import io.vertx.core.http.HttpServerOptions;
 import io.vertx.docgen.Source;
 import io.vertx.grpcio.server.GrpcIoServer;
 import io.vertx.grpcio.server.GrpcIoServiceBridge;
+import io.vertx.grpcio.server.GrpcServerIndex;
+import io.vertx.grpcio.server.ReflectionServiceV1Handler;
+
+import java.util.List;
 
 @Source
 public class GrpcIoServerExamples {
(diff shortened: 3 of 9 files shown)


vietj:
Putting that on radar

vertx-grpc-server/src/main/java/module-info.java
```diff
@@ -6,6 +6,7 @@
   requires io.vertx.codegen.api;
   requires io.netty.codec;
   requires io.netty.buffer;
+  requires com.google.protobuf;
```
vietj, line 9: that is actually fine since vertx-grpc-common already requires it we already need this dependency

vertx-grpc-common/src/main/java/io/vertx/grpc/common/ServiceMetadata.java
```diff
@@ -0,0 +1,91 @@
...
+ * and other metadata defined in the protobuf service definition.
+ */
+public interface ServiceMetadata {
+
+  /**
+   * Get the name of the service.
+   *
+   * @return the service name
+   */
+  ServiceName getServiceName();
```
vietj, line 24: instead `serviceName()`

vertx-grpc-common/src/main/java/io/vertx/grpc/common/ServiceMetadata.java
```diff
@@ -0,0 +1,91 @@
...
+
+  /**
+   * Get the service descriptor that contains detailed information about the service.
+   *
+   * @return the service descriptor
+   */
+  Descriptors.ServiceDescriptor getServiceDescriptor();
```
vietj, line 31: instead `serviceDescriptor()`

vertx-grpc-common/src/main/java/io/vertx/grpc/common/ServiceMetadata.java
```diff
@@ -0,0 +1,91 @@
...
+
+  /**
+   * Get a list of all method descriptors for this service.
+   *
+   * @return list of method descriptors
+   */
+  default List<Descriptors.MethodDescriptor> getMethods() {
```
vietj, line 38: instead `methodDescriptos()`

vertx-grpc-common/src/main/java/io/vertx/grpc/common/ServiceMetadata.java
```diff
@@ -0,0 +1,91 @@
...
+    return getServiceDescriptor().getMethods();
+  }
+
+  /**
+   * Get a method descriptor by name.
+   *
+   * @param methodName the name of the method
+   * @return an Optional containing the method descriptor if found, or empty if not found
+   */
+  default Optional<Descriptors.MethodDescriptor> getMethod(String methodName) {
```
vietj, line 48: instead `methodDescriptor(String)`

vertx-grpc-common/src/main/java/io/vertx/grpc/common/ServiceMetadata.java
```diff
@@ -0,0 +1,91 @@
...
+
+  /**
+   * Get the full path for a method, which can be used for making gRPC calls.
+   * The format is "/package.ServiceName/MethodName".
+   *
+   * @param methodName the name of the method
+   * @return the full path for the method
+   * @throws IllegalArgumentException if the method does not exist
+   */
+  default String getMethodPath(String methodName) {
```
vietj, line 72: instead `pathOfMethod`

vertx-grpc-common/src/main/java/io/vertx/grpc/common/ServiceMetadata.java
```diff
@@ -0,0 +1,91 @@
...
+    }
+    return getServiceName().pathOf(methodName);
+  }
+
+  /**
+   * Get a map of all methods in this service, with method names as keys and method descriptors as values.
+   *
+   * @return a map of method names to method descriptors
+   */
+  default Map<String, Descriptors.MethodDescriptor> getMethodsMap() {
```
vietj, line 84: I think we don't need that, if the demanding code needs it, it would build it

vertx-grpc-server/src/main/java/io/vertx/grpc/server/GrpcServer.java
```diff
@@ -86,4 +87,6 @@ static GrpcServer server(Vertx vertx, GrpcServerOptions options) {
   @GenIgnore(GenIgnore.PERMITTED_TYPE)
   <Req, Resp> GrpcServer callHandlerWithTranscoding(ServiceMethod<Req, Resp> serviceMethod, Handler<GrpcServerRequest<Req, Resp>> handler, MethodTranscodingOptions transcodingOptions);
 */
+
+  GrpcServer serviceMetadata(ServiceMetadata serviceMetadata);
```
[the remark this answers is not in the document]
vietj: or pehraps it's a method only there ?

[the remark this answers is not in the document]
vietj: Ok so I got a bit confused about the granularity of the description service (which works at the service level) versus the registration of call handlers in a gRPC server which works at the service method level. 

I think we get actually an usability issue here because registering service metadata is disconnected from handlers, so I think we should get an API that connects both with a builder style like. Let's take for example a service which has multiple methods like the streaming service.

Currently we have:

```java
server
  .serviceMetadata(STREAMING_SERVICE_METADATA)
  .callHandler(STREAMING_SOURCE, sourceHandler)
  .callHandler(STREAMING_PIPE, pipeHandler)
  .callHandler(STREAMING_SINK, sinkHandler)
```

I think instead we should have a notion of service that contains the metadata + the handlers

```java
Service service = Service
  .service(STREAMING_SERVICE)
  .descriptor(StreamingProto.getDescriptor().findServiceByName("Streaming"))
  .callHandler(STREAMING_SOURCE, sourceHandler)
  .build();

server.addService(service);
```

vietj (commented):
This looks good to me, as it stands I want to make the `Service` interface a read-only interface that only provides retrieval and binding of call handler and move the setter part / call handler to a builder instead that users would use.

As consequence the generated Vert.x XXService could simply implement the `Service` interface and provide the contract directly.

zZHorizonZz:
Yeah, I have been thinking about that as well, that the generated service could be extended by it. I will refactor that. 👍

vietj:
@zZHorizonZz thank you, I think we can merge soon, I will take care of further decoupling to a separate module

vietj:
thanks for the great contribution, here is the contribution follow-up https://github.com/eclipse-vertx/vertx-grpc/pull/141

PR #140, opened 2025-03-26
Supports JPMS to grpc io common

EmadAlblueshi:


EmadAlblueshi:
ping @vietj :-)

vietj:
unfortunately grpcio cannot really support JPMS to build modular application because it contains split packages (it might not be visible as is), so there is no JPMS support for grpcio jars

vietj:
that being said I think we could actually at least use automatic named modules at this stage

EmadAlblueshi:
@vietj Thank you for the clarifications.

PR #159, opened 2025-04-23
Fix health service example binding

zZHorizonZz:
Motivation:

Fix health service example binding

### DIFF
--- a/vertx-grpc-docs/src/main/java/examples/GrpcServerExamples.java
+++ b/vertx-grpc-docs/src/main/java/examples/GrpcServerExamples.java
@@ -326,8 +326,8 @@ public class GrpcServerExamples {
     // Register health checks for your services
     healthService.register("my.service.name", () -> Future.succeededFuture(true));
 
-    // Add the health service to the gRPC server
-    grpcServer.addService(healthService);
+    // Bind the health service to the server
+    healthService.bind(grpcServer);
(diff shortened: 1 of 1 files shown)


vertx-grpc-docs/src/main/java/examples/GrpcServerExamples.java
```diff
@@ -326,8 +326,8 @@ public void healthServiceExample(Vertx vertx, HttpServerOptions options) {
     // Register health checks for your services
     healthService.register("my.service.name", () -> Future.succeededFuture(true));
 
-    // Add the health service to the gRPC server
-    grpcServer.addService(healthService);
+    // Bind the health service to the server
+    healthService.bind(grpcServer);
```
vietj, line 330: I think it's better the way it is currently, what is the motivation for doing such change ?

zZHorizonZz: I wanted to make it the same as it is used in the vertx-examples repository, so that it is not confusing. I think one of these should possibly be fixed. I like addService better as well.

vietj: I think bind should not be used and most users should use addServices, if that's the other ay around we should change the other locations.

PR #163, opened 2025-04-24
Transcoding methods & generator CLI

zZHorizonZz:
Motivation:

This PR introduces updates to gRPC code generation, and improves CLI support.


vietj:
Actually we want to avoid generation of transcoding service method that uses the service method FQN as a path.

We want this feature but directly handled by the `TranscodingInvoker` just with a bare service method. That is, any `ServiceMethod` deployed in the server along with the `vertx-grpc-transcoding` dependency will create a GrpcInvocation with transcoding request for it.

Then we can remove the generation of the `Transcoding` inner class and in the `GrpcService` write either a transcoded service method when there is metadata or just a regular service method.

This simplifies the generated code and the user experience: the user only need to care about its gRPC service binding. This binding will bind all possible gRPC protocols for it.

Here is an example

Let's say we have a .proto file that contains two methods

- SayFoo and SayBar
- SayFoo has a /foo transcoding configuration
- SayBar is a regular service method

The generated `GreeterGrpcService` will have

```java
private static final ServiceMethod SayBar = ...;
private static final TranscodingServiceMethod SayFoo = ...;
```

The the user will simply do:

```java
grpcServer.addService(new GreeterGrpcService() {
  public FooReply sayFoo(FooRequest request) {
    // HTTP/2 application/grpc /Greeter/SayFoo
    // HTTP/2 application/grpc+json /Greeter/SayFoo
    // HTTP/x application/json /foo (via configuration)
  }
  public BarReply sayBar(BarRequest request) {
    // HTTP/2 application/grpc /Greeter/SayBar
    // HTTP/2 application/grpc+json /Greeter/SayBar
    // HTTP/x application/json /Greeter/SayBar (automatically)
  }
});
```

vertx-grpc-docs/src/main/java/examples/GrpcTranscodingExamples.java
```diff
@@ -10,7 +10,7 @@
 public class GrpcTranscodingExamples {
 
   public void transcodingRequestResponse(GrpcServer server) {
-    Service service = GreeterGrpcService.Transcoding.of(new GreeterGrpcService() {
+    Service service = GreeterGrpcService.of(new GreeterGrpcService() {
```
vietj, line 13: simply `new GreeterGrpcService() { }`, no need for `of()`

vertx-grpc-docs/src/main/java/examples/grpc/StreamingGrpcService.java
```diff
@@ -101,86 +101,43 @@ public static Service of(StreamingService service) {
     all.add(Source);
     all.add(Sink);
     all.add(Pipe);
+    all.add(SourceTranscodingMethod);
```
vietj, line 104: the generator should only create a single method whatsoever, that is `Source` we don't need repetition, we don't need to create two per service

vertx-grpc-protoc-plugin2/src/main/resources/grpc-service.mustache
```diff
@@ -87,62 +87,56 @@ public class {{className}} extends {{serviceName}}Service implements Service {
 {{#allMethods}}
     all.add({{methodName}});
 {{/allMethods}}
+{{#transcodingMethods}}
```
vietj, line 90: as said previously, instead of having:

- all methods
- all transcoding methods

we should have

- all methods, if a method declares transcoding meta data then the method is a transcoding method otherwise it is a regular method

Issue #166, opened 2025-04-27
Removing Vertx prefix from generated classes in vertx-grpc-protoc-plugin, cause duplicated class error


cristian-joseb:
### Questions

Do not use this issue tracker to ask questions, instead use one of these [channels](https://vertx.io/community/). Questions will likely be closed without notice.

### Version

Which version(s) did you encounter this bug ?
5.0.0.CR7

### Context

After the Vertx prefix was removed from the generated classes,  we have encountered duplicated class issue when using the vertx-grpc-protoc-plugin2 plugin .

### Do you have a reproducer?

A reproducer is a simple project hosted on GitHub (or another forge supporting git clone operation) that has a build file that can be executed to reproduce the issue.

Reproducers are very helpful for contributors and will likely help them fixing your bug faster.

* Link to github project/gist

### Steps to reproduce

1. Create a Vertx project using the ETCD proto files https://github.com/etcd-io/jetcd/tree/main/jetcd-core/src/main/proto
2. Run the mvn install.
3. The GRPC pluging will generated a Auth class under the generated/java directory, and Vert.x plugin will try to generate the same class under the generated/vertx directory.
4. build/generated/source/proto/main/vertx/io/etcd/jetcd/api/Auth.java:29: error: duplicate class: io.etcd.jetcd.api.Auth
  public interface Auth {

The issue seem to be related to generated Vert.x classes and files .

Maybe we can have an option to pass the service class name prefix or keep Vertx prefix as before.

### Extra

* Anything that can be relevant such as OS version, JVM version

PR #168, opened 2025-05-06
Generator service prefix

zZHorizonZz:
Motivation:

This change adds ability to specify service class prefix trough generator option --service-prefix, and moves generator options to options class for better readability.

Fixes: #166


vertx-grpc-docs/src/main/asciidoc/plugin.adoc
```diff
@@ -117,6 +117,7 @@ The plugin accepts the following arguments:
 - _--grpc-service_: generate the service files
 - _--grpc-io_: generate stub files compatible with `io.grpc:grpc-stub`, not generated by default
 - _--grpc-transcoding_: whether to generate transcoding options for methods with HTTP annotations
+- _--service-prefix_: generate service classes with a prefix
```
vietj, line 120: I think we need to explain that the original gRPC service name is prefixed to avoid confusion and that it will impact the generation of all classes, e.g. if prefix is Foo, then we would have FooGreeter, FooGreeterGrpcClient etc... and give an example in the doc so users understand easily

PR #172, opened 2025-05-09
feat: support protoc options

comynli:
Protobuf compiler typically pass arguments using -xxx_out=options:$path or -xxx_opt=options. This PR makes the vertx-grpc-protoc-plugin2 follow this convention


vietj:
we should be consistent and name like for the Maven plugin "grpc-io" and not "io", etc...

can you sign the Eclipse Contributor Agreement ?

comynli:
> we should be consistent and name like for the Maven plugin "grpc-io" and not "io", etc...
Ok, I'll fix it quickly.

vietj:
@comynli you need also to sign the eclipse contributor agreement so we can accept your contribution

comynli:
> @comynli you need also to sign the eclipse contributor agreement so we can accept your contribution

Already done. I am very sorry, I am not very familiar with English, and I spent some time understanding it.

vertx-grpc-docs/src/main/asciidoc/plugin.adoc
```diff
@@ -161,6 +161,51 @@ protobuf {
...
+    }
+    vertx {
+      artifact = "io.vertx:vertx-grpc-protoc-plugin2:${vertx.grpc.version}"
+    }
+  }
+  generateProtoTasks {
+    all()*.plugins {
+      grpc
+      vertx {
+        option("io")
```
vietj, line 195: should be updated with "grpc-io"

vietj:
can you instead reuse the same command line parser instead of parsing with a different syntax

zZHorizonZz:
Could we maybe use this type of options in IT test module instead of args to ensure everything works as expected? What do you think @vietj

vietj:
@zZHorizonZz I am not sure what you mean, can you elaborate ?

zZHorizonZz:
So I think we need to be able to test this somehow, because hoping that this won't break with potential updates etc. doesn't seem like a good idea. So I think we should at least be able to test it, and because you should be able to pass the plugin parameters through <pluginParameter></pluginParameter>, that should basically trigger this type of argument parsing?

zZHorizonZz:
I will close this as it has been superseded by rework of the plugin which does not relly on old jprotoc anymore.

PR #173, opened 2025-05-16
Transcoding optimazations

zZHorizonZz:
Motivation:

This PR optimizes some utility classes and added more test to path matcher test class.

### DIFF
--- a/vertx-grpc-transcoding/src/main/java/io/vertx/grpc/transcoding/impl/MessageWeaver.java
+++ b/vertx-grpc-transcoding/src/main/java/io/vertx/grpc/transcoding/impl/MessageWeaver.java
@@ -1,110 +1,137 @@
 package io.vertx.grpc.transcoding.impl;
 
 import io.vertx.core.buffer.Buffer;
-import io.vertx.core.internal.buffer.BufferInternal;
 import io.vertx.core.json.DecodeException;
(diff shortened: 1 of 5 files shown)


vietj (approved):

vietj:
I think we could write a JMH benchmark for this

zZHorizonZz:
I have been thinking about benchmarks for gRPC in general because there are TechEmpower tests which are testing Vert.x, but we don't really have a comparison of vertx-grpc with, for example, the grpc-ecosystem version. I don't think this is a major optimization because in general this only reduces some unnecessary calls, and in PercentEncoding it just puts the reserved chars to BitSet away from switch (which might make it a bit slower than using switch, but at the same time this way it is more readable, I think). Also if there was something for percent encoding already I would use that but I didn't really find anything in google lib it only provides escapers and not unescapers.

vietj:
here I am thinking about microbenchmarking the convertion without gRPC transport itself

zZHorizonZz:
Yeah, I can look into that :+1:

vietj:
we might be surprised by having extra bytes copy happening

zZHorizonZz:
Well, I have added some basic benchmarks (I used Vertx core for the setup) and I have run the tests with old and the new. For MessageWeaver class, the new version is faster, but PercentEncoding class is slower. I will try to look into that.

PR #176, opened 2025-05-21
Proper implementation of service error mapping to status mapping.

vietj:
Motivation:

The current mapping of errors to status is currently not homogeneous (between vertx/io service generation) and lacks of a status exception allowing to set force a status to be set by the wrapper.

Changes:

Introduce `io.vertx.grpc.server.StatusException` that conveys the grpc status to be returned by the service when this exception is encountered.

### DIFF
--- a/vertx-grpc-server/src/main/java/io/vertx/grpc/server/StatusException.java
+++ b/vertx-grpc-server/src/main/java/io/vertx/grpc/server/StatusException.java
@@ -1,0 +1,56 @@
+/*
+ * Copyright (c) 2011-2025 Contributors to the Eclipse Foundation
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0, or the Apache License, Version 2.0
+ * which is available at https://www.apache.org/licenses/LICENSE-2.0.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR Apache-2.0
+ */
+package io.vertx.grpc.server;
+
+import io.vertx.core.VertxException;
+import io.vertx.grpc.common.GrpcStatus;
+
+/**
+ * A glorified GOTO forcing a response status.
+ */
+public final class StatusException extends VertxException {
+
+  /**
+   * Map an exception to a GrpcStatus:
+   *
+   * <ul>
+   *   <li>{@link StatusException} returns {@link #status()}</li>
+   *   <li>{@link UnsupportedOperationException} returns {@link GrpcStatus#UNIMPLEMENTED}</li>
+   *   <li>otherwise returns {@link GrpcStatus#UNKNOWN}</li>
+   * </ul>
+   *
+   * @param t the exception to map
+   * @return the mapped status
+   */
+  public static GrpcStatus mapStatus(Throwable t) {
+    if (t instanceof StatusException) {
+      return ((StatusException)t).status();
+    } else if (t instanceof UnsupportedOperationException) {
+      return GrpcStatus.UNIMPLEMENTED;
+    } else {
+      return GrpcStatus.UNKNOWN;
+    }
+  }
+
+  private final GrpcStatus status;
+
+  public StatusException(GrpcStatus status) {
+    super("Grpc status " + status.name());
+    this.status = status;
+  }
+
+  /**
+   * @return the status
+   */
+  public GrpcStatus status() {
+    return status;
+  }
+}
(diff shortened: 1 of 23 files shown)


vertx-grpc-server/src/main/java/io/vertx/grpc/server/StatusException.java
```diff
@@ -0,0 +1,56 @@
...
+    } else if (t instanceof UnsupportedOperationException) {
+      return GrpcStatus.UNIMPLEMENTED;
+    } else {
+      return GrpcStatus.UNKNOWN;
+    }
+  }
+
+  private final GrpcStatus status;
+
+  public StatusException(GrpcStatus status) {
```
zZHorizonZz, line 45: Could we maybe have the ability to specify a status message? I think if you want to return Invalid Argument, sometimes you might want to specify exactly which argument is invalid. Or maybe could we create something like io.grpc status which contains code, description, and throwable (cause)?

vietj: that's a good point to add message, I missed that

vietj:
@zZHorizonZz please have a look now

vertx-grpc-server/src/main/java/io/vertx/grpc/server/impl/GrpcServerResponseImpl.java
```diff
@@ -24,7 +24,9 @@
 import io.vertx.grpc.common.impl.Utils;
 import io.vertx.grpc.server.GrpcProtocol;
 import io.vertx.grpc.server.GrpcServerResponse;
+import io.vertx.grpc.server.StatusException;
 
+import javax.swing.undo.StateEdit;
```
zZHorizonZz, line 29: Is this necessary?

vertx-grpc-server/src/main/java/module-info.java
```diff
@@ -10,8 +10,9 @@
   requires io.netty.codec;
   requires io.netty.buffer;
   requires com.google.protobuf;
+    requires java.desktop;
```
zZHorizonZz, line 13: I don't think this relates to the import of java swing

vertx-grpc-server/src/main/java/io/vertx/grpc/server/impl/GrpcServerResponseImpl.java
```diff
@@ -93,6 +95,17 @@ public void cancel() {
     }
   }
 
+  public void fail(Throwable failure) {
```
zZHorizonZz, line 98: Can't we have something for users to map their errors easily as well? I mean, sometimes I think they could call a library which might throw an exception, for example, but they would need to wrap the calls in try-catch to be able to handle the errors. I was thinking something like this [ExceptionHandlerProvider](https://github.com/quarkusio/quarkus/blob/3.23.0/extensions/grpc/api/src/main/java/io/quarkus/grpc/ExceptionHandlerProvider.java) might be useful for that?

vertx-grpc-server/src/main/java/io/vertx/grpc/server/StatusException.java
```diff
@@ -0,0 +1,49 @@
...
+
+import io.vertx.core.VertxException;
+import io.vertx.grpc.common.GrpcStatus;
+
+/**
+ * A glorified GOTO forcing a response status.
+ */
+public final class StatusException extends VertxException {
+
+  private final GrpcStatus status;
```
zZHorizonZz, line 21: Can't we have some unified object for the status with some nice builder methods, which might be used in GrpcServerResponseImpl instead of status and statusMessage? Also, I have noticed I don't think we are setting the statusMessage for transcoding response. I think we should maybe send it as a body?

vertx-grpc-server/src/main/java/io/vertx/grpc/server/impl/GrpcServerResponseImpl.java
vietj: nope, automatic IDE import :-)

vertx-grpc-server/src/main/java/module-info.java
vietj: automatic IDE import

vertx-grpc-server/src/main/java/io/vertx/grpc/server/impl/GrpcServerResponseImpl.java
vietj: no, because this is a low level API (compared to generated code), users using this will be able to figure out.

I think we could have such mapping in generated code with a method that can be overriden in the service class by the user.

vertx-grpc-server/src/main/java/io/vertx/grpc/server/StatusException.java
vietj: GrpcStatus is per the spec only a value, it does not have a message.

As for transcoding I think body is not a good idea because some users might want to decode it and fail application that expects another format (e.g. JSON), it could be conveyed as a response header though.

zZHorizonZz:
LGTM

PR #178, opened 2025-05-30
Decode/encode JSON wire format messages for grpc-io.

vietj:
Motivation:

The server grpc-io does not support properly transcoding requests due to lack of support for JSON in its decoders.

Changes:

Implement JSON support in grpc-io bridge message encoder/decoder. In addition reuse the bridge codec instead of using more or less the same code in the Utils class.


zZHorizonZz (approved):
Don't really see anything wrong, so LGTM

PR #194, opened 2025-07-18
Add support for gRPC error metadata propagation

adparts:
This PR introduces support for propagating structured gRPC error metadata from the server to the client across the Vert.x gRPC stack. It allows both client and server components to handle and transmit metadata in error responses consistently.

Motivation:
Handling rich error information in gRPC is critical for debugging, client behavior, and user feedback. Until now, only status codes and messages were propagated, with no support for custom trailers or metadata.

This change allows:
- Server-side exceptions to carry metadata.
- Middleware and clients to access that metadata.
- End-to-end visibility of error context for both internal and external clients.

This feature is required for internal use at our company, where we rely on Vert.x-based gRPC services and need to provide detailed error context (status, message, and metadata) from server to client. Without this functionality, we are unable to fully adopt vertx-grpc in our projects.

**Changes**
 
Server
Created GrpcErrorInfoProvider interface to allow exceptions to expose:

- A GrpcStatus
- A message
- Metadata (io.vertx.core.MultiMap)

Made existing StatusException implement GrpcErrorInfoProvider and adapt GrpcServerResponseImpl to return the metadata from GrpcErrorInfoProvider

Client
Updated InvalidStatusException to include a metadata field.
Modified GrpcClientResponseImpl and the gRPC proto plugin to populate the metadata field when the returned status is not GrpcStatus.OK.

Vert.x gRPC-IO Layer
Server: Inject metadata into StatusRuntimeException for correct transmission to clients.
Client: Parse metadata from headers if the response is trailersOnly, not just from trailers.

Improved (but still not ideal) detection of trailersOnly responses by checking if a payload has been read.

### DIFF
--- a/vertx-grpc-server/src/main/java/io/vertx/grpc/server/GrpcErrorInfoProvider.java
+++ b/vertx-grpc-server/src/main/java/io/vertx/grpc/server/GrpcErrorInfoProvider.java
@@ -1,0 +1,41 @@
+package io.vertx.grpc.server;
+
+import io.vertx.core.MultiMap;
+import io.vertx.grpc.common.GrpcStatus;
+
+/**
+ * Interface for providing detailed error information in gRPC server responses.
+ * <p>
+ * Implementing this interface allows exceptions to expose structured gRPC error details,
+ * including a status a descriptive error message, and optional trailers.
+ * </p>
+ * <p>
+ * This design enables custom exceptions to propagate meaningful and rich error context to gRPC clients
+ * without coupling to a specific exception class.
+ * </p>
+ */
+
+public interface GrpcErrorInfoProvider {
+
+  /**
+   * Returns the GrpcStatus associated with this error.
+   *
+   * @return the gRPC status
+   */
+  GrpcStatus status();
+
+  /**
+   * Returns the gRPC error message to send to the client.
+   *
+   * @return the error message as a string
+   */
+  String message();
+
+  /**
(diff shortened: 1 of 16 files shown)


vertx-grpc-server/src/main/java/io/vertx/grpc/server/GrpcErrorInfoProvider.java
```diff
@@ -0,0 +1,41 @@
...
+ * Implementing this interface allows exceptions to expose structured gRPC error details,
+ * including a status a descriptive error message, and optional trailers.
+ * </p>
+ * <p>
+ * This design enables custom exceptions to propagate meaningful and rich error context to gRPC clients
+ * without coupling to a specific exception class.
+ * </p>
+ */
+
+public interface GrpcErrorInfoProvider {
```
vietj, line 18: I don't think this interface is needed for implementing the feature ? if we can just use directly `StatusException` without it, then we should do it

vietj: Currently there are not tests using `GrpcErrorInfoProvider`, can you explain how this would be used in practice, it is not yet still clear to me how it can be used

vietj:
@adparts we would like actually to use an exception mapper strategy to transform an exception into a StatusException, so we make it less intrusive, this mapper would be set for now on the ServiceBuilder and used when deploying a service

PR #206, opened 2025-09-09
Refactor gRPC plugin implementation for modular architecture

zZHorizonZz:
Refactor the gRPC plugin by replacing `VertxGrpcGeneratorImpl` with a new modular architecture to improve maintainability, and allow future extendability.

This PR supersedes: #172


vertx-grpc-protoc-plugin2/src/main/java/io/vertx/grpc/plugin/protoc/ProtocRequestConverter.java
```diff
@@ -0,0 +1,174 @@
...
+  public static ProtocRequestConverter create(PluginProtos.CodeGeneratorRequest request) {
+    ProtobufTypeMapper typeMapper = new ProtobufTypeMapper(request.getProtoFileList());
+    return new ProtocRequestConverter(typeMapper);
+  }
+
+  public List<ServiceDescriptor> convertServices(PluginProtos.CodeGeneratorRequest request) {
+    List<ServiceDescriptor> services = new ArrayList<>();
+
+    for (DescriptorProtos.FileDescriptorProto fileProto : request.getProtoFileList()) {
+      if (!request.getFileToGenerateList().contains(fileProto.getName())) {
```
vietj, line 33: filtering should be done by the caller, that is the `VertxGrpcGenerator` should filter the list to be processed by the `ProtocRequestConverter` instead of relying on it to skip the files that should not be generated

vertx-grpc-protoc-plugin2/src/main/java/io/vertx/grpc/plugin/generation/generators/CodeGenerator.java
```diff
@@ -0,0 +1,43 @@
...
+
+/**
+ * Abstract base class for a code generator that produces generated files based on a specific type of generation context.
+ * <p>
+ * This class provides common functionalities such as file creation, relative path calculation, and integration with a template engine to facilitate rendering templates into
+ * generated files. Concrete implementations should extend this class and provide specific logic for the code generation process.
+ */
+public abstract class CodeGenerator {
+
+  protected final TemplateEngine templateEngine;
```
vietj, line 15: the CodeGenerator does not need to keep a reference to the template engine as we only have a single engine that is the mustache engine, this will simplify a bit.

vietj (changes requested):
more simplification :-)

PR #245, opened 2026-03-13
Decouple gRPC server implementation from the actual invocation layer

vietj:
Motivation:

The request/response API is coupled to the gRPC transport implementations leading to the usage of subclasses.

We would like to have this API independent of the actual transport for several reasons:

- simplify / untangle the interactions between the request/response API and the protocol semantics
- enable new transports, not HTTP based such as Vert.x event-bus
- enable transport level interception

Implementation:

- [x] decouple `GrpcServerResponse` implementation from invocation layer and HTTP
- [x] decouple `GrpcServerRequest` implementation from invocation layer and HTTP
- [x] simplify
- [x] decouple invocation layer from HTTP transport


vertx-grpc-server/src/main/java/io/vertx/grpc/server/impl/HttpGrpcInvoker.java
```diff
@@ -0,0 +1,178 @@
...
+      return writeMessage((GrpcMessageFrame) frame);
+    } else if (frame instanceof GrpcTrailersFrame) {
+      return writeTrailers((GrpcTrailersFrame) frame);
+    }
+    return contextInternal.failedFuture("Invalid message");
+  }
+
+  protected Future<Void> writeHeaders(GrpcHeadersFrame frame) {
+    MultiMap httpHeaders = httpResponse.headers();
+    httpHeaders.set("content-type", frame.contentType());
```
zZHorizonZz: Could we use HttpHeaders or HttpHeaderNames?

vertx-grpc-server/src/main/java/io/vertx/grpc/server/impl/GrpcServerResponseImpl.java
```diff
@@ -32,26 +33,26 @@
 /**
  * @author <a href="mailto:julien@julienviet.com">Julien Viet</a>
  */
-public abstract class GrpcServerResponseImpl<Req, Resp> extends GrpcWriteStreamBase<GrpcServerResponseImpl<Req, Resp>, Resp> implements GrpcServerResponse<Req, Resp> {
+public final class GrpcServerResponseImpl<Req, Resp> extends GrpcWriteStreamBase<GrpcServerResponseImpl<Req, Resp>, Resp> implements GrpcServerResponse<Req, Resp> {
 
   private static final Pattern COMMA_SEPARATOR = Pattern.compile(" *, *");
   private static final Set<String> GZIP_ACCEPT_ENCODING = Collections.singleton("gzip");
```
zZHorizonZz: Does this need to be Set? I mean is there any way that there could be different format of gzip? If so shouldn't this be just ACCEPT_ENCODING so that more encodings can be added in the future like snappy and such?

vertx-grpc-client/src/main/java/io/vertx/grpc/client/impl/SetIdleTimeoutFrame.java
```diff
@@ -0,0 +1,21 @@
...
+
+import java.time.Duration;
+
+public class SetIdleTimeoutFrame implements GrpcFrame {
+
+  private final Duration timeout;
+
+  public SetIdleTimeoutFrame(Duration timeout) {
+    if (timeout.isNegative()) {
+      throw new IllegalArgumentException("Timeout must be > 0");
```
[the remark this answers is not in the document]
vietj: isPositive only works with Java > 11

vietj:
merged elsewhere

PR #250, opened 2026-03-30
Expose callHandler registered services in service() method

### LINKED ISSUE #208: Reflection service doesn't include service methods
### Describe the feature

The reflection service 'list services' does not include services registered via 'addCallHandler', it only includes services registered via 'addService'
All of our services use a custom handler for each service method, so none of the services are listed on the server when using reflection.

### Contribution

_No response_

zZHorizonZz:
Motivation:

Services registered via GrpcIoServer.callHandler(MethodDescriptor, Handler) were not included in the reflection service's ListServices response. Only services registered via addService() appeared, because callHandler only populates the method routing map without adding a Service entry. This meant users who register all their service methods with custom handlers had no services visible through reflection.

Changes:

GrpcIoServerImpl now extracts the ProtoServiceDescriptorSupplier from the MethodDescriptor's schema descriptor when callHandler is called, and creates a Service entry for reflection. The services() method is overridden to merge these with addService registered services, deduplicating by name so the same service registered through both paths only appears once.

Fixes: #208

### DIFF
--- a/vertx-grpcio-server/src/main/java/io/vertx/grpcio/server/impl/GrpcIoServerImpl.java
+++ b/vertx-grpcio-server/src/main/java/io/vertx/grpcio/server/impl/GrpcIoServerImpl.java
@@ -10,22 +10,33 @@
  */
 package io.vertx.grpcio.server.impl;
 
+import com.google.protobuf.Descriptors;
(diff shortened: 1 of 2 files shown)


vietj (approved):
Looks good to me @zZHorizonZz

PR #252, opened 2026-04-14
Add option including JSON default value for transcoding

mutexd:
Motivation:

I'm using grpc-transcoding so that I can have one grpc-service implementation which can serve both Grpc/REST entry point. However, the JsonPrinter's default behavior excludes default values such as int32:0, string:"". Therefore, a GrpcResponse with 
`{ "message": "Hello", "status": 0 }`
will become like this for REST response.
 `{ "message": "Hello" }`
Even though it's valid in terms of ProtoJSON, it would be nice to have default values because JSON in REST conventionally includes default value like this.

Draft for #261 (need discussion)
Adds a flag for `vertx-grpc-protoc-plugin2` and directly embeds `printer.includingDefaultValueFields()` into transcoding's `encoder` method. Changing the `ServiceMethod` signature should be avoided and therefore I approach the issue from `vertx-grpc-protoc-plugin2`.

### DIFF
--- a/vertx-grpc-common/src/main/java/io/vertx/grpc/common/GrpcMessageEncoder.java
+++ b/vertx-grpc-common/src/main/java/io/vertx/grpc/common/GrpcMessageEncoder.java
@@ -17,6 +17,17 @@ public interface GrpcMessageEncoder<T> {
    */
   @GenIgnore
   static <T extends MessageLite> GrpcMessageEncoder<T> encoder() {
+    return encoder(JsonFormat.printer());
+  }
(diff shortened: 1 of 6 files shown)


mutexd:
If the approach taken doesn't align with the structure of the project, I would like to delete this PR and create an issue. A feedback would be appreciated.

zZHorizonZz:
Hello, thank you for the contribution.

I think it would be better to convert this to a draft and open an issue so it can be discussed, more in depth.

PR #257, opened 2026-04-16
Add gRPC EventBus transport support

zZHorizonZz:
Motivation:
                                                                                                                                                                                                                                 
  - Added a new gRPC transport that uses the Vert.x EventBus instead of HTTP/2 for local or clustered communication within a Vert.x application
                                                                                                                                                                                                                                 
  Changes:
                                                                                                                                                                                                                                 
  - Added a new vertx-grpc-event-bus module with dedicated client (EventBusGrpcClient) and server (EventBusGrpcServer) APIs                                                                                                      
  - The server reuses GrpcServerRequestImpl/GrpcServerResponseImpl from the server module, so generated service stubs work out of the box via addService


vertx-grpc-client/src/main/java/io/vertx/grpc/client/GrpcClientService.java
```diff
@@ -0,0 +1,27 @@
...
+
+/**
+ * A contract for issuing gRPC client requests by service method, without requiring
+ * a specific transport or server address.
+ *
+ * <p>Generated gRPC client stubs use this interface as their underlying transport abstraction,
+ * allowing them to work with any implementation - HTTP/2 via {@link GrpcClient}, event bus, etc.</p>
+ */
+@VertxGen
+public interface GrpcClientService {
```
vietj, line 16: why isn't GrpcClient extending that ?

zZHorizonZz: I forgot that one, thanks for pointing it :+1:

PR #258, opened 2026-04-17
Correctly handle HttpServerResponse reset signals.

vietj:
Motivation:

Vert.x gRPC server does not handle HttpServerResponse reset signals, leading to missing signal from the gRPC interaction perspective.

Changes:

Also handle gRPC server response.


vietj:
cc @anavarr

PR #265, opened 2026-04-29
Fix exception in decoding empty message

mutexd:
Motivation:

- Http::Get request (the request type is empty message) gets 400 error instead of 200 because the empty request body cannot be decoded into EmptyMessage Type.
- GrpcMessageDecoder:decode failed to decode the message and CodecException is thrown for message type without any property

Solution:
- The MessageWeaver.weaveRequestMessage used to return transcoded=`{}` but now return transcoded= `""`, which results in decode exception.
- Revert the check.
- One unit test is added for Http::Get with request type as empty message.

### DIFF
--- a/vertx-grpc-it/src/test/java/io/vertx/grpc/it/TranscodingTest.java
+++ b/vertx-grpc-it/src/test/java/io/vertx/grpc/it/TranscodingTest.java
@@ -276,28 +276,6 @@ public class TranscodingTest extends ProxyTestBase {
   }
 
   @Test
-  public void testUnaryWithEmptyMessage() throws TimeoutException {
-    HttpClient client = vertx.createHttpClient();
-    vertx.createHttpServer()
-      .requestHandler(GrpcServer.server(vertx).callHandler(GreeterGrpcService.SayHelloEmpty, call -> call.handler(helloRequest -> {
-        HelloReply helloReply = HelloReply.newBuilder().setMessage("Hello there").build();
-        call.response().end(helloReply);
-      }))).listen(8080, "localhost").await(10, TimeUnit.SECONDS);
-
-  RequestOptions options = new RequestOptions().setHost("localhost").setPort(8080).setURI("/v1/hello").setMethod(HttpMethod.GET);
-
-  Buffer body = client.request(options).compose(req -> {
-      req.putHeader(HttpHeaders.CONTENT_TYPE, "application/json");
-      req.putHeader(HttpHeaders.ACCEPT, "application/json");
-      return req.send("");
-    }).expecting(HttpResponseExpectation.SC_OK)
-    .expecting(HttpResponseExpectation.JSON)
-    .compose(HttpClientResponse::body)
-    .await(10, TimeUnit.SECONDS);
-    assertEquals("Hello there", getMessage(body.toString()));
-  }
-
-  @Test
   public void testIoUnaryWithoutOption1() throws TimeoutException {
     testUnaryWithoutOption(server -> server.callHandler(GreeterGrpc.getSayHelloWithoutOptionsMethod(), call -> call.handler(helloRequest -> {
       HelloReply helloReply = HelloReply.newBuilder().setMessage("Hello " + helloRequest.getName()).build();
--- a/vertx-grpc-it/src/test/java/io/vertx/grpc/it/TranscodingTest.java
+++ b/vertx-grpc-it/src/test/java/io/vertx/grpc/it/TranscodingTest.java
@@ -276,6 +276,28 @@ public class TranscodingTest extends ProxyTestBase {
   }
 
   @Test
+  public void testUnaryWithEmptyMessage() throws TimeoutException {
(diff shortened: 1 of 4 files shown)


vertx-grpc-it/src/test/proto/helloworld.proto
```diff
@@ -116,3 +123,7 @@ message HelloBodyRequest {
 message HelloBodyResponse {
   HelloReply response = 1;
 }
+
+// The request message containing nothing
+message HelloEmptyMessage {
```
zZHorizonZz: Could we use google.protobuf.Empty message for this?

mutexd: Sure. I noticed that there's already a `empty.proto` file. I imported that one and changed it to `grpc.test.Empty`.

zZHorizonZz:
Alright, @mutexd thank you it seem to be passing just to be sure can you also rebase? Once that is done I think I can approve a merge it today in evening or tomorrow in the morning, so it should be then available in vertx 5.1 which is planned for this month.

vertx-grpc-it/src/test/java/io/vertx/grpc/it/TranscodingTest.java
zZHorizonZz, line 1: Could we move test out of TranscodingTest in it module and instead test this in ServerTranscodingTest directly inside of transcoding module?

zZHorizonZz (changes requested):

vertx-grpc-it/src/test/java/io/vertx/grpc/it/TranscodingTest.java
mutexd: Done.

vertx-grpc-transcoding/src/test/java/io/vertx/tests/transcoding/ServerTranscodingTest.java
```diff
@@ -111,6 +113,17 @@ public void setUp(TestContext should) {
       copyTrailers(request.headers(), request.response().trailers());
       request.response().end(Empty.newBuilder().build());
     });
+    grpcServer.callHandler(UNARY_EMPTY_REQUEST_CALL, request -> {
```
zZHorizonZz: We should fix the current empty handler. See the review comment for more information.

vertx-grpc-transcoding/src/test/java/io/vertx/tests/transcoding/ServerTranscodingTest.java
```diff
@@ -256,6 +269,26 @@ public void testEmpty(TestContext should) {
       }));
   }
 
+  @Test
+  public void testUnaryEmptyRequest(TestContext should) {
```
zZHorizonZz: We should not need this test as testEmpty is supposed to test this but handler is setup in a way it is not catching this.

zZHorizonZz (commented):
OK so I did more digging into the matter why we did not catch it with empty call before. Basically the decoding is never run if we never set handler, that is why empty call was actually passing. I think we can simplify this pr by just fixing the original test that was supposed to catch this.

So basically handler should look like this:
```java
grpcServer.callHandler(EMPTY_CALL, request -> {
      request.handler(requestMsg -> {
        copyHeaders(request.headers(), request.response().headers());
        copyTrailers(request.headers(), request.response().trailers());
        request.response().end(Empty.newBuilder().build());
      });
    });
```

But instead it looks like this at the moment:
```java
grpcServer.callHandler(EMPTY_CALL, request -> {
      copyHeaders(request.headers(), request.response().headers());
      copyTrailers(request.headers(), request.response().trailers());
      request.response().end(Empty.newBuilder().build());
    });
```

mutexd:
Setting the handler doesn't suffice it. You need to `setBody("")` in order to reproduce the issue.
I modified it accordingly. However, I have to point out that moving test and tinkering the original test don't make sense here. The bug is reproducible in an integration test scenario then it should be kept there. Making it happen in  ServerTranscodingTest baffles people why the test is setup like that.

zZHorizonZz:
Thanks for digging this up. After more time on it, looking at the HttpRule spec, the C++ reference impl in grpc-httpjson-transcoding, and Microsoft's gRPC JSON transcoding in ASP.NET Core, the changes needed are bigger than what's here. The slow path has the same loose pass through, the body vs bindings ordering should match what the spec implies (the body comment in http.proto describes `*` as mapping fields not captured by the path pattern, so path bindings should be authoritative).

Happy to either merge this as-is for the immediate 400 fix and follow up with another PR for the conformance changes, or push the broader fix directly into your branch. Whichever works better for you.

mutexd:
Thanks. I do not fully follow what's needed here. I would suggest we merge this and you could create another PR afterward.

zZHorizonZz (approved):

zZHorizonZz:
LGTM

Thank you, I will do the follow up PR then with changes

PR #266, opened 2026-05-04
Add support for WireFormat configuration

### LINKED ISSUE #261: Add option including JSON default value for transcoding
### Describe the feature

The nice thing about grpc-transcoding is that one can have one grpc-service implementation which can serve both Grpc/REST entry point. However, the JsonPrinter's default behavior excludes default values such as int32:0, string:"". Therefore, a GrpcResponse with
{ "message": "Hello", "status": 0 }
will become like this for REST response.
{ "message": "Hello" }
Even though it's valid in terms of ProtoJSON, it would be nice to have default values because JSON in REST conventionally includes default value like this.

Suggestion: Add an option in build time or even run time for including JsonDefaultValue in grpc-transcoding

### Contribution

_No response_

zZHorizonZz:
Motivation:

- Allow customization of JSON printer and parser configurations for gRPC transcoding trough WireFormat configuration.

Changes:

- Added ability to configure supported wireformats on server.
- Added `JsonWireFormat` and `ProtobufWireFormat` as concrete implementations of `WireFormat`.
- Updated `GrpcMessageEncoder` and `GrpcMessageDecoder` to respect custom wire formats.
- Added `ProtobufJsonReader` and `ProtobufJsonWriter`.
- Updated transcoding logic to use specific wire formats.

Closes: #261 
Supersedes: #252


zZHorizonZz:
@mutexd If you want you can also review, I think this should allow more configuration in the future

vertx-grpc-common/src/main/java/io/vertx/grpc/common/JsonWireFormat.java
```diff
@@ -0,0 +1,221 @@
...
+/**
+ * JSON {@link WireFormat}. Carries {@link WriterConfig} and {@link ReaderConfig} that
+ * protobuf-aware encoders and decoders use to drive {@link ProtobufJsonWriter} and
+ * {@link ProtobufJsonReader}.
+ * <p>
+ * Instances are immutable. Use {@link #withWriterConfig(WriterConfig)} and
+ * {@link #withReaderConfig(ReaderConfig)} to derive a configured copy.
+ */
+@DataObject
+public class JsonWireFormat implements WireFormat {
```
vietj, line 18: I think we don't need to have WriterConfig/ReaderConfig.

Instead the state should be on JsonWireFormat itself and have JsonWireFormat immutable.

users would do in practice: `WireFormat.JSON.alwaysPrintFieldsWithNoPresence(true)`

mutexd:
> @mutexd If you want you can also review, I think this should allow more configuration in the future

It looks good to me. I also tried it out and it works just as expected. Thanks.

vertx-grpc-client/src/main/java/io/vertx/grpc/client/impl/Http2GrpcOutboundStream.java
```diff
@@ -97,15 +100,13 @@ private Future<Void> handleHeadersFrame(GrpcHeadersFrame frame, boolean end) {
...
-      case PROTOBUF:
-        contentType = "application/grpc";
-        break;
-      case JSON:
-        contentType = "application/grpc+json";
-        break;
-      default:
-        throw new UnsupportedOperationException();
+    WireFormat format = frame.format();
+    if (format instanceof ProtobufWireFormat) {
```
vietj: I think WireFormat should have a mediaType() method that returns the mime type for the format

vertx-grpc-common/src/main/java/io/vertx/grpc/common/JsonWireFormat.java
```diff
@@ -0,0 +1,198 @@
...
+   * Default {@code sortingMapKeys} flag = {@code false}.
+   */
+  public static final boolean DEFAULT_SORTING_MAP_KEYS = false;
+
+  /**
+   * Default {@code ignoringUnknownFields} flag = {@code false}.
+   */
+  public static final boolean DEFAULT_IGNORING_UNKNOWN_FIELDS = false;
+
+  private final boolean alwaysPrintFieldsWithNoPresence;
```
vietj: implementation should be a single byte instead of 8 booleans using masking.

this way we can lazy cache format in a 2^6 array (lazy populated) when they are created and avoid creating un-necessary instance of wireformat.

vietj:
we should move ProtobufJsonReader and ProtobufJsonWriter to impl package and not expose this for the moment, and we open it to module that actually use it, I want to keep the surface API small

vertx-grpc-common/src/main/java/io/vertx/grpc/common/impl/ProtobufJsonReader.java
```diff
@@ -0,0 +1,31 @@
...
+import io.vertx.grpc.common.CodecException;
+import io.vertx.grpc.common.JsonWireFormat;
+
+/**
+ * Reads JSON {@link Buffer} payloads into a protobuf {@link Message.Builder}.
+ * <p>
+ * Use {@link #create(JsonWireFormat)} to get an instance. The default implementation is backed
+ * by {@code com.google.protobuf.util.JsonFormat}.
+ */
+public interface ProtobufJsonReader {
```
vietj: I think for now we can keep ProtobufJsonReader and ProtobufJsonReadImpl a single class

vertx-grpc-common/src/main/java/io/vertx/grpc/common/impl/ProtobufJsonWriter.java
```diff
@@ -0,0 +1,31 @@
...
+import io.vertx.grpc.common.CodecException;
+import io.vertx.grpc.common.JsonWireFormat;
+
+/**
+ * Writes a protobuf {@link MessageOrBuilder} as a JSON {@link Buffer}.
+ * <p>
+ * Use {@link #create(JsonWireFormat)} to get an instance. The default implementation is backed
+ * by {@code com.google.protobuf.util.JsonFormat}.
+ */
+public interface ProtobufJsonWriter {
```
vietj: I think for now we can keep ProtobufJsonWriter and ProtobufJsonWriterImpl a single class

vertx-grpc-server/src/main/java/io/vertx/grpc/server/impl/GrpcServerImpl.java
```diff
@@ -151,12 +151,22 @@ private int validate(GrpcServerRequestInspector.RequestInspectionDetails details
       return 415;
     }
 
+    if (!options.isFormatEnabled(details.format)) {
```
vietj, line 154: is that case tested ?

zZHorizonZz: It is vertx-grpc-server/src/test/java/io/vertx/tests/server/FormatSupportTest.java

vietj (approved):

PR #268, opened 2026-05-07
Fix REST endpoint dosn't work in GrpcTrancoding when the url path contains verb

### LINKED ISSUE #277: REST endpoint doesn't work in GrpcTranscoding for Url with verb
### Version

5.0.10

### Context

For url with verb (such as `/path/foo/bar:run`), the grpc client can make the request and get response successfully, but a rest client gets error.

### Steps to reproduce

1. define a grpc call and bind it to a url path with verb (`/path/foo/bar:run`)
2. make a call with a grpc-client -> works
3. make a call with a web-cliet -> error

### Do you have a reproducer?

_No response_

mutexd:
Motivation:

This PR tries to address the issue #277 

For rest api url: `/path/foo/bar:run`, the registered call handler will not be called because it's registered as `/path/foo/bar`. When the web-client makes the call, the server tries to search for `/path/foo/bar:run`.

The grpc endpoint works fine because the handler is registered by it's package path name and also searched by the package name.

Solution:

In `TranscodingServiceMethodImpl:computePaths`, `/path/foo/bar:run` is registered to `/path/foo/bar` thus later on the match couldn't be found. We should append the verb since HttpTemplate actually parses it.

### DIFF
--- a/vertx-grpc-transcoding/src/main/java/io/vertx/grpc/transcoding/impl/TranscodingServiceMethodImpl.java
+++ b/vertx-grpc-transcoding/src/main/java/io/vertx/grpc/transcoding/impl/TranscodingServiceMethodImpl.java
@@ -71,6 +71,10 @@ public class TranscodingServiceMethodImpl<I, O> implements TranscodingServiceMet
       }
       sb.append('/').append(a);
     }
+    String verb = tmpl.getVerb();
+    if (verb != null && !verb.isEmpty()) {
+      sb.append(':').append(verb);
+    }
     paths.add(sb.toString());
     List<MethodTranscodingOptions> extra = options.getAdditionalBindings();
     if (extra != null) {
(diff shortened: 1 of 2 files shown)


zZHorizonZz:
Hey, sorry for not responding to this, we have been busy preparing the 5.1 release. I had a quick glance and it seems to be ok. I will have better look tomorrow if I find some time, although it will probably need to be rebased properly as there seem to be conflicts.

zZHorizonZz (approved):
Looks good, after rebase I think it will be ready for merge

mutexd:
I might misunderstand your comment. Should I set target branch as 5.1?

zZHorizonZz:
> I might misunderstand your comment. Should I set target branch as 5.1?

No keep the main there will be 5.1.2 sometime in the  near future so I think we can just backport it after in separate pr. If you need this in 5.1, otherwise it will be available in 5.2 snapshots.

zZHorizonZz:
So I think github has issue with the merge commits and doesn't allow rebase because of that, I have push permissions so I can fix it if you want and cleanup the history for you or if you want you can do it yourself.

mutexd:
I guess it has something to do with the signing-key I was using. I fixed that one now.

zZHorizonZz:
So I got around to attempt to fix it, basically the issue is the merge commits which cannot be used for rebase and merge on github. I have cleanup the thing and pushed it to my fork so if you want you can take it from me and push it to your branch and sign it with your key though I think it will show me as coauthor. Basically for clean rebase we need to get rid of the merge commits.

https://github.com/zZHorizonZz/vertx-grpc/commits/bugfix/verb-in-rest-path-is-truncated/

mutexd:
@zZHorizonZz You could directly overwrite the commits in my PR.

zZHorizonZz:
Alright I will do that sometime later today. If everything goes trough I will hopefully rebase and merge it to main branch 👍

zZHorizonZz:
@mutexd can you check this, if everything is as it should be and nothing is missing?

mutexd:
@zZHorizonZz It's all there. Thanks.

zZHorizonZz:
Thank you for your contribution!

PR #269, opened 2026-05-11
Transcoding/spec conformant message weaver

zZHorizonZz:
## Motivation

Follow-up to #265. The original PR fixed the 400 on empty-message RPCs by adjusting the fast-track and adding `setBody("")` to the test, but the underlying `MessageWeaver` was not actually conforming to the `google.api.HttpRule` body semantics:

- When `body` was unset, the HTTP request body was still being merged at the root of the gRPC message, even though per the spec an unset `body` means the HTTP body should be ignored.
- When `body = "*"`, a shallow merge was used and path/query bindings did not consistently take precedence on collision.
- A latent bug in `TranscodingServiceMethodImpl.accept` always read the body field from the top-level `MethodTranscodingOptions`, silently ignoring any `body` declared on an `additional_bindings` entry.

This PR makes the behavior match the HttpRule contract, to align with the reference implementations in C++ ([grpc-ecosystem/grpc-httpjson-transcoding `request_weaver.h`](https://github.com/grpc-ecosystem/grpc-httpjson-transcoding/blob/master/src/include/grpc_transcoding/request_weaver.h)) and C# ([Microsoft.AspNetCore.Grpc.JsonTranscoding `JsonRequestHelpers.ReadMessage`](https://github.com/dotnet/aspnetcore/blob/main/src/Grpc/JsonTranscoding/src/Microsoft.AspNetCore.Grpc.JsonTranscoding/Internal/JsonRequestHelpers.cs)). The spec itself is the `body` field comment in [`google/api/http.proto`](https://github.com/googleapis/googleapis/blob/master/google/api/http.proto).

### DIFF
--- a/vertx-grpc-docs/src/main/asciidoc/transcoding.adoc
+++ b/vertx-grpc-docs/src/main/asciidoc/transcoding.adoc
@@ -115,6 +115,7 @@ service Greeter {
       get: "/v1/hello/{name}"
       additional_bindings {
         post: "/v1/hello"  // Alternative POST endpoint
+        body: "*"          // Use the HTTP body as the request message
       }
     };
   }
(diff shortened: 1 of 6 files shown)


zZHorizonZz:
@mutexd This is follow up of your PR. You can check it and review it if you want

PR #271, opened 2026-05-12
Redefine the role of the io.vertx.grpc.server.Service interface.

vietj:
Motivation:

The `Service` interface acts for now as a IoC contract for its implementation to configure a` GrpcServer`.

This implies that the `Service` interface is intimate with a server.

Instead we can decouple it from the server contract and make it handle service calls.

Changes:

`Service` does not bind anymore to the server, instead it handles service calls.

The default server implementation now delegates calls to the service contract.


vietj:
need to fix a bug before merging

PR #281, opened 2026-06-09
Add gRPC streaming over the event bus

zZHorizonZz:
Motivation:

Add gRPC streaming over the event bus


vertx-grpc-eventbus/src/main/java/io/vertx/grpc/eventbus/impl/EventBusGrpcClientStreamingCall.java
```diff
@@ -0,0 +1,249 @@
...
+
+    if (timeout != null) {
+      options.setSendTimeout(timeout.toMillis());
+    }
+
+    if (requestHeaders != null) {
+      EventBusHeaders.encodeMultiMap(HEADER_PREFIX, requestHeaders, options.getHeaders());
+    }
+
+    long clientStreamId = endpoint.register(this);
```
vietj: we should split registration and stream ID generation in two steps, so we register only when the interaction on the following EventBus request is successful

we should have a monotonic stream ID sequence that is allocated when the stream is created and sent when the result of the event-bus interaction is received

vertx-grpc-eventbus/src/main/proto/io/vertx/grpc/eventbus/transport/v1alpha/eventbus_transport.proto
```diff
@@ -0,0 +1,78 @@
...
+
+  oneof frame {
+    Initialize initialize = 3; // client to server, the opening request body
+    Initialized initialized = 4; // server to client, the open reply
+    Message message = 5; // a message payload, either direction
+    WindowUpdate window_update = 6; // flow-control credit, either direction
+    HalfClose half_close = 7; // client to server, end of the request stream
+    Trailers trailers = 8; // server to client, terminates the call
+    Cancel cancel = 9; // either direction, abnormal termination
+    Headers headers = 10; // server to client, response metadata, before the first message
```
vietj: I think we don't need this frame because we can make this part of the initial request/response interaction along with the initialized response in clear format (like for unary)

vertx-grpc-eventbus/src/main/proto/io/vertx/grpc/eventbus/transport/v1alpha/eventbus_transport.proto
```diff
@@ -0,0 +1,78 @@
...
+// opens with a request (its body an Initialize) whose reply is an Initialized. Each
+// endpoint multiplexes all of its streams over a single private address, and
+// stream_id demuxes the call on that address. Frames are serialized as protobuf
+// binary. The handshake and flow control are described in the module README.
+message TransportFrame {
+  uint64 stream_id = 1; // the destination endpoint's id for this call, demuxes it on the shared private address
+  uint64 stream_sequence = 2; // per-stream, monotonic, advances on Message frames only
+
+  oneof frame {
+    Initialize initialize = 3; // client to server, the opening request body
```
vietj: I don't think we need initialize/initialized frames here, we should keep protobuf frames only for the private address communication

vertx-grpc-eventbus/src/main/java/io/vertx/grpc/eventbus/impl/EventBusStreamEndpoint.java
```diff
@@ -0,0 +1,93 @@
...
+    return address;
+  }
+
+  long register(FrameHandler stream) {
+    long id = sequence.incrementAndGet();
+    streams.put(id, stream);
+    return id;
+  }
+
+  void closed(FrameHandler stream) {
```
vietj: are we testing that an an initial or streaming failure correctly remove the stream from the map ?

vietj (changes requested):

vertx-grpc-eventbus/src/main/java/io/vertx/grpc/eventbus/impl/FrameHandler.java
```diff
@@ -0,0 +1,9 @@
+package io.vertx.grpc.eventbus.impl;
+
+import io.vertx.core.Closeable;
+import io.vertx.core.eventbus.Message;
+import io.vertx.grpc.eventbus.transport.v1alpha.TransportFrame;
+
+interface FrameHandler extends Closeable {
```
vietj: for the moment we don't need that interface, we should just make in an abstract method of EventBusStreamEndpoint

vertx-grpc-eventbus/src/main/java/io/vertx/grpc/eventbus/impl/EventBusStreamEndpoint.java
```diff
@@ -0,0 +1,107 @@
...
+import java.util.List;
+import java.util.UUID;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
+import java.util.concurrent.atomic.AtomicLong;
+
+abstract class EventBusStreamEndpoint {
+
+  protected final ConcurrentMap<Long, FrameHandler> streams = new ConcurrentHashMap<>();
+  protected volatile int maxConcurrentStreams = EventBusGrpcServerOptions.DEFAULT_MAX_CONCURRENT_STREAMS;
```
vietj: it should be removed.

I don't think we can support this because there is no way to identify the remote side we will handle when using the event-bus due to load balancing, the main purpose of max concurrent streams is to protect the producer and when using load balancing we just use an event-bus address.

vertx-grpc-common/src/main/java/io/vertx/grpc/common/ServiceMethod.java
```diff
@@ -19,6 +19,10 @@
 public interface ServiceMethod<I, O> {
 
   static <Req, Resp> ServiceMethod<Resp, Req> client(ServiceName serviceName, String methodName, GrpcMessageEncoder<Req> encoder, GrpcMessageDecoder<Resp> decoder) {
+    return client(serviceName, methodName, MethodType.UNARY, encoder, decoder);
```
vietj: we should set null instead of MethodType.UNARY instead because this is not available and unknown. I'd prefer a bug with an NPE rather than a misbehavior.

vertx-grpc-common/src/main/java/io/vertx/grpc/common/ServiceMethod.java
```diff
@@ -19,6 +19,10 @@
...
+  }
+
+  static <Req, Resp> ServiceMethod<Resp, Req> client(ServiceName serviceName, String methodName, MethodType type, GrpcMessageEncoder<Req> encoder, GrpcMessageDecoder<Resp> decoder) {
```
vietj: instead of this we should instead have booleain indicating cardinality for both sides of the service and avoid expose MethodType at the moment, it should remain internal.

vietj (changes requested):

vertx-grpc-common/src/main/java/io/vertx/grpc/common/MethodType.java
```diff
@@ -0,0 +1,58 @@
+package io.vertx.grpc.common;
+
+/**
+ * The cardinality of a gRPC service method.
+ */
+public enum MethodType {
```
vietj: I think for now I would like to avoid introducing this type as it is only used in this module and instead stick with the two booleans used, probably those booleans should be Boolean to indicate when the information is unknown

vertx-grpc-eventbus/src/main/java/io/vertx/grpc/eventbus/impl/EventBusGrpcClientStreamingCall.java
```diff
@@ -0,0 +1,277 @@
...
+import static io.vertx.grpc.eventbus.impl.EventBusHeaders.HEADER_PREFIX;
+import static io.vertx.grpc.eventbus.impl.EventBusHeaders.TRAILER_PREFIX;
+
+class EventBusGrpcClientStreamingCall extends EventBusGrpcStreamBase {
+
+  private final EventBusStreamEndpoint endpoint;
+  private final EventBus eventBus;
+  private final ServiceName serviceName;
+  private final String methodName;
+  private final Deque<GrpcMessage> pending = new ArrayDeque<>();
```
vietj: init that in the constructor along with other inits

vertx-grpc-eventbus/src/main/java/io/vertx/grpc/eventbus/impl/EventBusGrpcClientStreamingCall.java
```diff
@@ -0,0 +1,277 @@
...
+
+  private WireFormat wireFormat;
+  private String encoding;
+  private MultiMap requestHeaders;
+  private Duration timeout;
+
+  private boolean ended;
+  private State state = State.IDLE;
```
vietj: init that in the constructor along with other inits

vertx-grpc-eventbus/src/main/java/io/vertx/grpc/eventbus/impl/EventBusGrpcStreamBase.java
```diff
@@ -1,18 +1,130 @@
...
+    if (flowing && granted <= window / 2) {
+      int delta = window - granted;
+      if (delta > 0) {
+        granted += delta;
+        sendTransportFrame(TransportFrame.newBuilder().setWindowUpdate(WindowUpdate.newBuilder().setDelta(delta)));
+      }
+    }
+  }
+
+  protected void writeMessage(GrpcMessage message) {
```
vietj: We need to add message write future return with the following modification:

the `EventBusGrpcStreamBase#writeMessage(GrpcMessage)` should return a `Future<Void>`, to implement this:

1. the `EventBusGrpcStreamBase#outboundQueue` should be modified to contain an object that has a Promise in addition of the message itself, e.g. a `MessageWrite` internal class
2. the method `EventBusGrpcStreamBase#sendTransportFrame` should be modified to return a `Future<Void>`

The implementation of `sendTransportFrame` should use a `MessageProducer` instead of the event bus directly, `MessageProducer#write` will return a `Future<Void>`, this producer for sending can be obtained with `EventBus#sender(String address)`, before writing the message delivery options should be set on the producer (and then set to null). I think this is OK but not convenient, so I think we should overload `MessageProducer#write(T)` with `MessageProducer#write(T, DeliveryOptions)` in vertx-core at some point.

vietj (changes requested):

vietj (changes requested):
A few changes are needed:

1. we need to test event bus failures impact on the stream (discussed offline)
2. `EventBusGrpcServer` should not extend `Closeable` which is rather used for internal purpose, instead it should define its own close() method like an `HttpServer`.
3. `EventBusGrpcClient` should have a close() method like an `HttpClient`.

vertx-grpc-eventbus/src/main/java/io/vertx/grpc/eventbus/impl/EventBusGrpcServerStreamingCall.java
```diff
@@ -0,0 +1,218 @@
...
+        if (clientListening) {
+          sendResponseHeaders(responseHeaders);
+        } else {
+          pendingHeaders = responseHeaders;
+          headersPending = true;
+        }
+        return context.succeededFuture();
+      case MESSAGE:
+        return writeMessage(((GrpcMessageFrame) frame).message());
+      case TRAILERS:
```
vietj, line 109: this should also return a future that returns the outcome of the trailers write

vertx-grpc-eventbus/src/main/java/io/vertx/grpc/eventbus/impl/EventBusGrpcServerStreamingCall.java
```diff
@@ -0,0 +1,218 @@
...
+        enqueue(new TrailersWrite((GrpcTrailersFrame) frame));
+        return context.succeededFuture();
+      default:
+        return context.succeededFuture();
```
vietj: I think this should return a failed future as writing an incorrect frame should not be permited by default

vertx-grpc-eventbus/src/main/java/io/vertx/grpc/eventbus/impl/EventBusGrpcServerStreamingCall.java
```diff
@@ -0,0 +1,218 @@
...
+    switch (frame.type()) {
+      case HEADERS:
+        MultiMap responseHeaders = ((GrpcHeadersFrame) frame).headers();
+        if (clientListening) {
+          sendResponseHeaders(responseHeaders);
+        } else {
+          pendingHeaders = responseHeaders;
+          headersPending = true;
+        }
+        return context.succeededFuture();
```
vietj: we should here also return a future that is not always succeeded, e.g. sendResponseHeaders could definitely return a future that we would return here

vertx-grpc-eventbus/src/main/java/io/vertx/grpc/eventbus/impl/EventBusGrpcServerStreamingCall.java
```diff
@@ -0,0 +1,218 @@
...
+  }
+
+  @Override
+  public Future<Void> end(GrpcFrame frame) {
+    return write(frame).compose(v -> end());
+  }
+
+  @Override
+  public Future<Void> end() {
+    return context.succeededFuture();
```
vietj: I think when we write the last grpc frame we should keep a reference to the future returned by this call and return it here (and if it is null we should somehow thrown an illegal state exception ?), that's because the end future should be completed after all the other frames have been sent

vertx-grpc-eventbus/src/main/java/io/vertx/grpc/eventbus/impl/EventBusGrpcClientStreamingCall.java
```diff
@@ -0,0 +1,324 @@
...
+        encoding = headersFrame.encoding();
+        requestHeaders = headersFrame.headers();
+        timeout = headersFrame.timeout();
+        return open();
+      case MESSAGE:
+        return onMessageWrite(((GrpcMessageFrame) frame).message());
+      case CANCEL:
+        return sendCancel();
+      default:
+        return context.succeededFuture();
```
vietj: we should here return instead a failed future to signal the stream cannot interpret the frame

vietj (changes requested):
I think we need to have a default ping timeout > 0 to ensure reliability since the event bus does not provide a connection oriented medium, we cannot have zero ping timeout and zero ping interval

vertx-grpc-eventbus/src/main/java/io/vertx/grpc/eventbus/impl/MessageWrite.java
```diff
@@ -0,0 +1,37 @@
...
+ * Implementing classes provide the specific behavior for message writing and error handling based on the context they are used in.
+ */
+interface MessageWrite {
+
+  /**
+   * Indicates whether the message exchange uses a windowed flow-control mechanism.
+   *
+   * @return {@code true} if the message exchange utilizes windowed flow control; {@code false} otherwise.
+   */
+  default boolean windowed() {
```
vietj: rather name it flowControlled

vertx-grpc-eventbus/src/main/java/io/vertx/grpc/eventbus/EventBusGrpcServerOptions.java
```diff
@@ -0,0 +1,124 @@
...
+   */
+  public static final Set<WireFormat> DEFAULT_SUPPORTED_WIRE_FORMATS = Collections.unmodifiableSet(new LinkedHashSet<>(Arrays.asList(WireFormat.PROTOBUF, WireFormat.JSON)));
+
+  /**
+   * The default maximum ping interval in milliseconds = {@code 120_000} (2 minutes)
+   */
+  public static final long DEFAULT_MAX_PING_INTERVAL = 120_000L;
+
+  private Set<WireFormat> supportedWireFormats;
+  private long maxPingInterval;
```
vietj: we should use java.time.Duration instead of long

vertx-grpc-eventbus/src/main/java/io/vertx/grpc/eventbus/EventBusGrpcClientOptions.java
```diff
@@ -0,0 +1,137 @@
...
+   */
+  public static final long DEFAULT_PING_INTERVAL = 30_000L;
+
+  /**
+   * The default ping timeout in milliseconds = {@code 60_000} (60 seconds), twice the default ping interval
+   */
+  public static final long DEFAULT_PING_TIMEOUT = 60_000L;
+
+  private WireFormat wireFormat;
+  private long pingInterval;
```
vietj: we should use java.time.Duration instead of long

vertx-grpc-eventbus/src/main/java/io/vertx/grpc/eventbus/EventBusGrpcClientOptions.java
```diff
@@ -0,0 +1,137 @@
...
+  private long pingTimeout;
```
vietj: we should use java.time.Duration instead of long

vertx-grpc-common/src/main/java/io/vertx/grpc/common/impl/GrpcReadStreamBase.java
```diff
@@ -52,6 +55,10 @@ public Buffer payload() {
   private final GrpcMessageDecoder<T> messageDecoder;
   private final Promise<Void> end;
   private Handler<GrpcError> errorHandler;
+  private final Deque<GrpcMessage> pending = new ArrayDeque<>();
```
vietj: we should avoid changing this file, I have ran tests with the _main_ version of `GrpcReadStreamBase` and they actually pass, so we should revert this file unless we can exhibit a case that makes a difference

vertx-grpc-eventbus/src/main/java/io/vertx/grpc/eventbus/EventBusGrpcClientOptions.java
```diff
@@ -0,0 +1,137 @@
...
+      throw new IllegalArgumentException("pingTimeout must be greater than 0");
+    }
+    this.pingTimeout = pingTimeout;
+    return this;
+  }
+
+  /**
+   * @return a JSON representation of options
+   */
+  public JsonObject toJson() {
```
vietj: we should not provide this, that's really not useful and we tend to get away from those

vertx-grpc-eventbus/src/main/java/io/vertx/grpc/eventbus/EventBusGrpcClientOptions.java
```diff
@@ -0,0 +1,137 @@
...
+    JsonObject json = new JsonObject();
+    EventBusGrpcClientOptionsConverter.toJson(this, json);
+    return json;
+  }
+
+  @Override
+  public String toString() {
```
vietj: not needed

vertx-grpc-eventbus/src/main/java/io/vertx/grpc/eventbus/EventBusGrpcClientOptions.java
```diff
@@ -0,0 +1,137 @@
...
+  public EventBusGrpcClientOptions(EventBusGrpcClientOptions other) {
+    wireFormat = other.wireFormat;
+    pingInterval = other.pingInterval;
+    pingTimeout = other.pingTimeout;
+  }
+
+  /**
+   * Creates options from JSON.
+   */
+  public EventBusGrpcClientOptions(JsonObject json) {
```
vietj: we should not provide this, that's really not useful and we tend to get away from those

vertx-grpc-eventbus/src/main/java/io/vertx/grpc/eventbus/EventBusGrpcServerOptions.java
```diff
@@ -0,0 +1,124 @@
...
+   */
+  public EventBusGrpcServerOptions(EventBusGrpcServerOptions other) {
+    supportedWireFormats = new LinkedHashSet<>(other.supportedWireFormats);
+    maxPingInterval = other.maxPingInterval;
+  }
+
+  /**
+   * Creates options from JSON.
+   */
+  public EventBusGrpcServerOptions(JsonObject json) {
```
vietj: we should not provide this, that's really not useful and we tend to get away from those

vertx-grpc-eventbus/src/main/java/io/vertx/grpc/eventbus/EventBusGrpcServerOptions.java
```diff
@@ -0,0 +1,124 @@
...
+      throw new IllegalArgumentException("maxPingInterval must be greater than 0");
+    }
+    this.maxPingInterval = maxPingInterval;
+    return this;
+  }
+
+  /**
+   * @return a JSON representation of options
+   */
+  public JsonObject toJson() {
```
vietj: we should not provide this, that's really not useful and we tend to get away from those

vertx-grpc-eventbus/src/main/java/io/vertx/grpc/eventbus/EventBusGrpcServerOptions.java
```diff
@@ -0,0 +1,124 @@
...
+    JsonObject json = new JsonObject();
+    EventBusGrpcServerOptionsConverter.toJson(this, json);
+    return json;
+  }
+
+  @Override
+  public String toString() {
```
vietj: not needed

vertx-grpc-eventbus/src/main/java/io/vertx/grpc/eventbus/EventBusGrpcServerOptions.java
```diff
@@ -0,0 +1,124 @@
...
+import io.vertx.core.json.JsonObject;
+import io.vertx.grpc.common.WireFormat;
+
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.LinkedHashSet;
+import java.util.Set;
+
+@DataObject
+@JsonGen(publicConverter = false)
```
vietj: no json gen needed

vertx-grpc-eventbus/src/main/java/io/vertx/grpc/eventbus/EventBusGrpcClientOptions.java
```diff
@@ -0,0 +1,137 @@
+package io.vertx.grpc.eventbus;
+
+import io.vertx.codegen.annotations.DataObject;
+import io.vertx.codegen.annotations.Unstable;
+import io.vertx.codegen.json.annotations.JsonGen;
+import io.vertx.core.json.JsonObject;
+import io.vertx.grpc.common.WireFormat;
+
+@DataObject
+@JsonGen(publicConverter = false)
```
vietj: no json gen needed

vietj (changes requested):

vertx-grpc-eventbus/README.md
```diff
@@ -0,0 +1,510 @@
...
+equivalent to an HTTP upgrade. The delivery headers contain the data, as for a unary
+call. The body of the request and the body of the reply are empty.
+
+#### The request of the client
+
+The client sends the first `request()` call to the service address. The headers contain
+these items:
+
+- `action`, the name of the method.
+- `grpc-wire-format`, the wire format.
```
vietj, line 82: I think this list is not accurate anymore since headers have been changed / updated

vertx-grpc-eventbus/src/main/java/io/vertx/grpc/eventbus/impl/EventBusStreamEndpoint.java
```diff
@@ -0,0 +1,259 @@
...
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Set;
+import java.util.UUID;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
+import java.util.concurrent.TimeoutException;
+import java.util.concurrent.atomic.AtomicLong;
+
+abstract class EventBusStreamEndpoint {
```
vietj: I think this should be renamed `EventBusGrpcEndpoint` as sub classes handle both streaming and unaries

vertx-grpc-eventbus/src/main/java/io/vertx/grpc/eventbus/EventBusGrpcClient.java
```diff
@@ -26,24 +26,51 @@
...
+   * private address is bound.
    *
    * @param vertx the vertx instance
    * @param eventBus the event bus to use as transport
-   * @return the created client
+   * @return a future of the created client
    */
-  static EventBusGrpcClient client(Vertx vertx, EventBus eventBus) {
-    return new EventBusGrpcClientImpl(vertx, eventBus);
+  static Future<EventBusGrpcClient> client(Vertx vertx, EventBus eventBus) {
```
vietj: vertx has a single event bus, so this overload is not needed.

vertx-grpc-eventbus/src/main/java/io/vertx/grpc/eventbus/EventBusGrpcClient.java
```diff
@@ -26,24 +26,51 @@
...
+  /**
+   * Create an event bus gRPC client using the provided event bus and options. The returned future completes once the
+   * client's private address is bound.
+   *
+   * @param vertx the vertx instance
+   * @param eventBus the event bus to use as transport
+   * @param options the client options
+   * @return a future of the created client
+   */
+  static Future<EventBusGrpcClient> client(Vertx vertx, EventBus eventBus, EventBusGrpcClientOptions options) {
```
vietj: vertx has a single event bus, so this overload is not needed.

vertx-grpc-eventbus/src/main/java/io/vertx/grpc/eventbus/EventBusGrpcServer.java
```diff
@@ -23,34 +23,68 @@
...
+   * private address is bound.
    *
    * @param vertx the vertx instance
    * @param eventBus the event bus to use as transport
-   * @return the created server
+   * @return a future of the created server
    */
-  static EventBusGrpcServer server(Vertx vertx, EventBus eventBus) {
-    return new EventBusGrpcServerImpl(vertx, eventBus);
+  static Future<EventBusGrpcServer> server(Vertx vertx, EventBus eventBus) {
```
vietj: vertx has a single event bus, so this overload is not needed.

vertx-grpc-eventbus/src/main/java/io/vertx/grpc/eventbus/EventBusGrpcServer.java
```diff
@@ -23,34 +23,68 @@
...
+  /**
+   * Create an event bus gRPC server using the provided event bus and options. The returned future completes once the
+   * server's private address is bound.
+   *
+   * @param vertx the vertx instance
+   * @param eventBus the event bus to use as transport
+   * @param options the server options
+   * @return a future of the created server
+   */
+  static Future<EventBusGrpcServer> server(Vertx vertx, EventBus eventBus, EventBusGrpcServerOptions options) {
```
vietj: vertx has a single event bus, so this overload is not needed.

vertx-grpc-eventbus/src/main/java/io/vertx/grpc/eventbus/EventBusGrpcServerOptions.java
```diff
@@ -0,0 +1,100 @@
...
+
+import java.time.Duration;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.LinkedHashSet;
+import java.util.Set;
+
+@DataObject
+@Unstable
+public class EventBusGrpcServerOptions {
```
vietj, line 18: Javadoc

vietj (changes requested):
Minor changes :-)

vietj (approved):

PR #285, opened 2026-07-09
Avoid quadratic buffer reassembly in GrpcReadStreamBase.handle

krickert:
`handle()` compacts the accumulated buffer on every arriving HTTP/2 DATA frame, even when no complete message was consumed (`idx == 0`). `Buffer.getBuffer()` copies, so while a message larger than one frame accumulates, every frame re-copies everything received so far — reassembly is O(n²) in message size, on the event loop. With 16 KB frames, an 8 MB message costs ~2 GB of memcpy and one full-size allocation per frame, and every other stream multiplexed on the same connection stalls behind it.

This change compacts only after at least one message was consumed, so each byte is copied at most twice (append + one compaction) and an accumulating message is left in place.

Measured on a unary service whose handler does identical linear work per byte (p50, single stream, plaintext localhost, window large enough that the sender never stalls on flow control):

| message size | before | after |
|---|---|---|
| 1 MB | 30.4 ms | 22.8 ms |
| 2 MB | 63.8 ms | 39.5 ms |
| 4 MB | 174.3 ms | 68.7 ms |
| 8 MB | 585.8 ms | 138.5 ms |

Before: latency grows 2.1×, 2.7×, 3.4× per size doubling (approaching the quadratic 4×). After: linear in message size.

Applies to the 4.x line (present in 4.5.28 and 4.5.29). The 5.x deframer (`Http2GrpcMessageDeframer`) accumulates in place and does not have the quadratic path.

### DIFF
--- a/vertx-grpc-common/src/main/java/io/vertx/grpc/common/impl/GrpcReadStreamBase.java
+++ b/vertx-grpc-common/src/main/java/io/vertx/grpc/common/impl/GrpcReadStreamBase.java
@@ -162,10 +162,15 @@ public abstract class GrpcReadStreamBase<S extends GrpcReadStreamBase<S, T>, T> 
     if (pause) {
       stream.pause();
     }
-    if (idx < buffer.length()) {
-      buffer = buffer.getBuffer(idx, buffer.length());
-    } else {
+    if (idx == buffer.length()) {
       buffer = null;
+    } else if (idx > 0) {
+      // Only compact after at least one message was consumed. Compacting
+      // unconditionally re-copied the entire accumulated buffer on every
+      // arriving frame while a message larger than one HTTP/2 frame was
+      // still accumulating (idx == 0), making reassembly O(n^2) in the
+      // message size.
+      buffer = buffer.getBuffer(idx, buffer.length());
     }
   }
 


vietj:
@krickert can you sign the eclipse contributor agreement ?

krickert:
"You successfully submitted an Eclipse Contributor Agreement."

I keep signing it.  It says I have a bad email. It's a good email and matches my commit name at that one google account.  It says on my profile that I signed it though.  But  this thing doesn't allow me to verify - I'm logged in and I keep being told that my name isn't in the db.

{"status_code":500,"message":"Internal server error while processing request: Cannot invoke \"String.equals(Object)\" because the return value of \"java.net.URI.getHost()\" is null","url":null}

I also got this once :)

Hoepfully it's not from a large packet over grpc!

krickert:
OK I signed away.  I hate backoffice tools.  I think we're good to go.

krickert:
Thanks for merging - I have been chasing this one.  Can you cut a new point release or ensure that this is in the latest SNAPSHOT?  I'm about to push a build and I would need to maintain a custom jar -not super hard though... just a step I might forget to revert

vietj:
there shall be a release early next week, it should be in snapshot but I haven't verified

PR #287, opened 2026-07-17
Fix REST endpoint dosn't work in GrpcTrancoding when the url path contains verb (backport from main)

mutexd:
Motivation:

I would like to backport the fix on main branch to 5.1.x. ([Issue](https://github.com/eclipse-vertx/vertx-grpc/issues/277) | [Pull Request](https://github.com/eclipse-vertx/vertx-grpc/pull/268))
As I want to start to use 5.1.x and I would like to have this fix also included in 5.1.x.

Thanks for the help.


zZHorizonZz:
LGTM, Thank you

PR #310, opened 2026-08-20
HTTP grpc reset handling sends a cancel frame only when an exception handler is set.

vietj:


vietj:
not tested ATM, will be when testing invokers will be done.

PR #322, opened 2026-08-25
add support for DEADLINE_EXCEEDED

anavarr:
replaces https://github.com/eclipse-vertx/vertx-grpc/pull/267

Add support for DEADLINE_EXCEEDED status when a deadline expires

**Motivation:**
When a gRPC call's deadline expires, the client should report a `DEADLINE_EXCEEDED` status to the caller.
Previously, the `VertxClientCall` did not check the deadline on stream completion, so a timed-out call could end up being reported with a misleading status (e.g. `CANCELLED` or whatever the underlying transport error mapped to). 

**Modifications:**
In `VertxClientCall`, on the `grpcResponse.end()` callback, we now check if the deadline has expired before processing the normal response status. 
This check is applied in both the success path (response ended cleanly but after the deadline) and the failure path (response ended with an error). 
In both cases, if the deadline is expired, `DEADLINE_EXCEEDED` is returned immediately instead of propagating the underlying transport error.

`ClientBridgeTest` now asserts `DEADLINE_EXCEEDED` instead of `CANCELLED`

**Result:**
gRPC calls that time out now correctly throw `Status.DEADLINE_EXCEEDED` instead of `CANCELLED`

### DIFF
--- a/vertx-grpcio-client/src/main/java/io/vertx/grpcio/client/VertxClientCall.java
+++ b/vertx-grpcio-client/src/main/java/io/vertx/grpcio/client/VertxClientCall.java
@@ -127,6 +127,10 @@ class VertxClientCall<RequestT, ResponseT> extends ClientCall<RequestT, Response
             }
             readAdapter.init(grpcResponse, decoder);
             grpcResponse.end().onComplete(ar -> {
+              if (deadline != null && deadline.isExpired()) {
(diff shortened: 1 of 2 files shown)


vietj:
@anavarr can you udpate the commit message and pull request to follow the format described here https://github.com/eclipse-vertx/vert.x/blob/master/CONTRIBUTING.md#commit-message

vietj:
thanks @anavarr can you backport to 5.1 branch ?

PR #343, opened 2026-09-01
Change mapping of NO_ERROR to cancelled

anavarr:
**Motivation**
Issue https://github.com/quarkusio/quarkus/issues/56268 in Quarkus indicates that Vert.x and Netty treat `RST_STREAM` `NO_ERROR` differently: Vert.x assigned it `GrpcError.Internal` because of switch case-fallthrough, whereas Netty mapped it as a cancellation.

**Changes**
This PR changes NO_ERROR mapping to treat it as a Cancellation.


vietj:
This is not in agreement with the gRPC over HTTP/2 specification that says that 0 should be mapped to internal (https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md).

PR #355, opened 2026-09-04
Let the generator generate FQN identifier expressions instead of relying on class imports

### LINKED ISSUE #351: [code-gen] Service called `Service` in .proto generates class with import problems
### Version

5.1.7

### Context

While migrating from Vert.x 4 to Vert.x 5 I encountered an error in generated gRPC sources while using the vertx-grpc-protoc-plugin2 (version 5.1.7):

Image

The problems seems to be a naming conflict when generating from a protobuf file which defines a service called `Service`.

- from the contract template a Java interface `Service` is created
- from the service template a Java class `ServiceService` is created, which:
  - imports `io.vertx.grpc.server.Service` 
  - wants to implement the previously created interface `Service`
- this causes `ServiceService` to think it has to implement  `io.vertx.grpc.server.Service` instead of the generated interface

### Steps to reproduce

1. simple protobuf with service called `Service`:

```proto
syntax = "proto3";

option java_multiple_files = true;
option java_package = "examples";
option java_outer_classname = "HelloWorldProto";
package helloworld;

// The Service service definition.
service Service {
  // Sends a greeting
  rpc SayHello (HelloRequest) returns (HelloReply) {}
}

// The request message containing the user's name.
message HelloRequest {
  string name = 1;
}

// The response message containing the greetings
message HelloReply {
  string message = 1;
}
```

2. use protobuf generator plugin:
 
```xml
 <plugin>
                <groupId>io.github.ascopes</groupId>
                <artifactId>protobuf-maven-plugin</artifactId>
                <version>5.1.8</version>
                <configuration>
                    <protoc>4.29.3</protoc>
                    <plugins>
                        <plugin kind="binary-maven">
                            <groupId>io.grpc</groupId>
                            <artifactId>protoc-gen-grpc-java</artifactId>
                            <version>1.65.0</version>
                        </plugin>

                        <plugin kind="jvm-maven">
                            <groupId>io.vertx</groupId>
                            <artifactId>vertx-grpc-protoc-plugin2</artifactId>
                            <version>5.1.7</version>
                            <options>grpc-client=true,grpc-service=true,grpc-io=true</options>
                        </plugin>
                    </plugins>
                </configuration>
                <executions>
                    <execution>
                        <id>compile-java</id>
                        <goals>
                            <goal>generate</goal>
                        </goals>
                    </execution>
                </executions>
            </plugin>
```

3. run `mvn clean protobuf:generate`
4. look at generated class `ServiceService`

### Do you have a reproducer?

_No response_

zZHorizonZz:
Motivation:

The generator gives the contract interface the name of the service. A service with the name `Service` thus gets a contract interface `Service`. The generated `ServiceService` class imports `io.vertx.grpc.server.Service`. In Java, an import hides a type of the same name in the same package. Thus the class implements the wrong `Service`, and the code does not compile. This occurs for each service that has the name of a type that the templates import. `Future` is another example.

Changes:

- Each template declares its own type with `serviceNameBase`.
- Each template refers to the other generated types with the full names in `contractFqn`, `clientFqn`, `serviceFqn` and `outerFqn`. These fields did not contain full names before.
- The fields `grpcClientFqn`, `grpcServiceFqn` and `grpcIoFqn` are removed. Each of them spelled only its own declaration.
- The contract template has no imports. It declares the bare service name, thus an import of the same name is a compile error. The template gives the full names of the two types that it uses.
- The other templates lose their unused imports. `io.vertx.grpc.server.Service` in the service template is the import of the report.

Results:

The generated code does not rely anymore on imports and minimize conflicts with user provided names.

Closes #351

### DIFF
--- a/vertx-grpc-protoc-plugin2/src/main/java/io/vertx/grpc/plugin/generation/context/ServiceTemplateContext.java
+++ b/vertx-grpc-protoc-plugin2/src/main/java/io/vertx/grpc/plugin/generation/context/ServiceTemplateContext.java
@@ -12,14 +12,11 @@ public class ServiceTemplateContext {
   public String javaPackageFqn;
   public String serviceName;
   public String packageName;
+  public String serviceNameBase;
   public String contractFqn;
   public String serviceFqn;
   public String clientFqn;
-  public String grpcClientFqn;
-  public String grpcServiceFqn;
-  public String grpcIoFqn;
   public String outerFqn;
-  public String prefixedServiceName;
   public boolean codegenEnabled;
   public String javaDoc;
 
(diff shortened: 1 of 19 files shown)


vertx-grpc-protoc-plugin2/src/main/java/io/vertx/grpc/plugin/generation/context/ServiceTemplateContext.java
```diff
@@ -12,14 +12,11 @@ public class ServiceTemplateContext {
   public String javaPackageFqn;
   public String serviceName;
   public String packageName;
+  public String prefixedServiceName;
```
vietj: name this rather `serviceNameBase`

vietj (changes requested):
Minor change

vietj (approved):

PR #357, opened 2026-09-09
Build the gRPC-Web interop image from a local Dockerfile

zZHorizonZz:
Motivation:

The gRPC-Web interop job builds grpc-web's `net/grpc/gateway/docker/prereqs/Dockerfile` from the upstream checkout. Both stages of that file target Debian 11, whose release file is expired, so `apt-get update` fails with exit code 100 on the first `RUN` and the job never reaches the tests.

```
E: Release file for http://deb.debian.org/debian-security/dists/bullseye-security/InRelease
   is expired (invalid since 19h 5min 45s)
```

Reported upstream as grpc/grpc-web#1554, with the fix proposed in grpc/grpc-web#1553.

Changes:

- The job builds a copy of that Dockerfile kept in `vertx-grpc-server/src/test/docker/grpc-web-interop`. It differs from the original, https://github.com/grpc/grpc-web/blob/7c572794c45b91f096fe06fbbafbb5411612a414/net/grpc/gateway/docker/prereqs/Dockerfile, in versions alone.
- `InteropITest` builds the same Dockerfile when no prebuilt image is given.

Results:

The interop job builds its image again.

### DIFF
--- a/.github/workflows/ci-matrix-5.x.yml
+++ b/.github/workflows/ci-matrix-5.x.yml
@@ -48,7 +48,7 @@ jobs:
       # (moby/moby#33974). Testcontainers can't drive BuildKit itself
       # (testcontainers/testcontainers-java#2857), so we prebuild and pass the tag.
       - name: Build gRPC-Web interop image
-        run: DOCKER_BUILDKIT=1 docker build -t grpcweb-interop:ci -f _grpc-web/net/grpc/gateway/docker/prereqs/Dockerfile _grpc-web
+        run: DOCKER_BUILDKIT=1 docker build -t grpcweb-interop:ci -f vertx-grpc-server/src/test/docker/grpc-web-interop/Dockerfile _grpc-web
       - name: Run tests
         run: mvn -s .github/maven-ci-settings.xml -q clean verify -B -pl :vertx-grpc-server -am -Dgrpc-web.repo.path="$GITHUB_WORKSPACE/_grpc-web" -Dgrpc-web.image=grpcweb-interop:ci
   Deploy:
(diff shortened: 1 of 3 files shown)


.github/workflows/ci-matrix-5.x.yml
```diff
@@ -33,8 +33,8 @@ jobs:
       - name: Checkout gRPC-Web
         uses: actions/checkout@v6
         with:
-          repository: grpc/grpc-web
-          ref: master
+          repository: zZHorizonZz/grpc-web
+          ref: update-docker-toolchain
```
tsegismont: We shouldn't depend on a repository that is out of the stack, I think

tsegismont (changes requested):
Thanks for sharing @zZHorizonZz 

I think we should either have a fork of the gRPC Web repo in the Vert.x org, or we should script the update and run it in a fresh checkout during the build.

.github/workflows/ci-matrix-5.x.yml
zZHorizonZz: Alright I will change that I have 2 commits first one has the Dockerfile it was just one of the options how to handle this.

tsegismont:
@zZHorizonZz have you filed an issue with grpc-web ? I think they might accept a PR if you explain the problem

zZHorizonZz:
@tsegismont I did create PR in grpc-web it is referenced in description though I will probably create issue as well.

zZHorizonZz:
Alright the issue is at https://github.com/grpc/grpc-web/issues/1554

zZHorizonZz:
@tsegismont So the question is if you want this to be fork I dont have issue with it being fork underneath the vertx though somebody would need to do it or we can use local dockerfile for now and switch it back once (if) the pr is merged.

tsegismont:
Yes, we can use a customized Dockerfile here until, hopefully, your PR is merged in grpc-web repo

vietj (approved):
LGTM

PR #358, opened 2026-09-09
Fix REST url path with verb and param

mutexd:
Motivation:

The REST calll for callHandler registered at url="/task/{param}:run" (having verb **and** param) doesn't work in 5.1.6
I investigated this and it turned out that this following fix caused this side effect.
[Regist](https://github.com/eclipse-vertx/vertx-grpc/pull/287)

url=`/task:run` is stored to mountPath=`/task:run` so it works fine.
url=`/task/{param}:run` is stored to mountPath=`/task:run` and therefore when the request comes, grpc-transcoding-server cannot find the handler. It should be stored to mountPath=`/task/` and when grpc-trancoding-server iterating through registered handlers, the correct one can return success.

Thus, I suggest we avoid appending verb to the mount path when there's variable and verb in the url at the same time.
This also happens on main. I'll port it to main if it's approved.

### DIFF
--- a/vertx-grpc-transcoding/src/test/java/io/vertx/tests/transcoding/ServerTranscodingTest.java
+++ b/vertx-grpc-transcoding/src/test/java/io/vertx/tests/transcoding/ServerTranscodingTest.java
@@ -368,7 +368,7 @@ public class ServerTranscodingTest extends GrpcTestBase {
   public void testPathParamWithVerb(TestContext should) {
     String payload = "foobar";
     httpClient.request(HttpMethod.POST, "/task/" + payload + ":run").compose(req -> {
-      String body = encode(EchoRequest.newBuilder().setPayload(payload).build()).toString();
+      String body = encode(EchoRequest.newBuilder().build()).toString();
       req.headers().addAll(HEADERS);
       return req.send(body).compose(response -> response.body().map(response));
     }).onComplete(should.asyncAssertSuccess(response -> should.verify(v -> {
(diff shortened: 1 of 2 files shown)


vertx-grpc-transcoding/src/test/java/io/vertx/tests/transcoding/ServerTranscodingTest.java
```diff
@@ -348,8 +365,23 @@ public void testVerbWithBody(TestContext should) {
   }
 
   @Test
-  public void testPayloadWithBody(TestContext should) {
+  public void testPathParamWithVerb(TestContext should) {
     String payload = "foobar";
+    httpClient.request(HttpMethod.POST, "/task/" + payload + ":run").compose(req -> {
+      String body = encode(EchoRequest.newBuilder().setPayload(payload).build()).toString();
```
zZHorizonZz: If the routing is working correctly then we should not need to set the payload because payload should be properly derived from the path.

zZHorizonZz (changes requested):

vertx-grpc-transcoding/src/test/java/io/vertx/tests/transcoding/ServerTranscodingTest.java
mutexd: You're right. I removed it.

zZHorizonZz (approved):
LGTM

zZHorizonZz:
@mutexd would it be possible to open the PR for main branch as well?

PR #370, opened 2026-09-16
Fix gRPC server not cancelling call on RST_STREAM NO_ERROR from peer

anavarr:
**Motivation**
see https://github.com/quarkusio/quarkus/issues/56268
When a peer (e.g. Envoy) sends RST_STREAM with NO_ERROR (0x0), the cancellation path was silently skipped: `GrpcWriteStreamBase.handleError()` only called `handleCancel()` for GrpcError.CANCELLED, and so did `GrpcIoServiceBridgeImpl` error handler.
The handler ran to completion, wrote a late response to a dead stream, and destroyed the entire HTTP/2 connection (thus killing all other multiplexed streams on it).

**Changes**
- Removed the GrpcError.CANCELLED guard in `GrpcWriteStreamBase.handleError()` so handleCancel() is called for any transport error
- Removed the equivalent guard in `GrpcIoServiceBridgeImpl` error handler so listener.onCancel() is invoked for any error on a still-open stream.

The gRPC-over-HTTP/2 spec maps RST_STREAM NO_ERROR to INTERNAL rather than CANCELLED, so propagating it through listener.onCancel() stack might seem wrong but the spec also specifies "immediate full-closure" for any RST_STREAM, and grpc-java `NettyServerHandler` already calls onCancel() for all client-sent RST_STREAM regardless of error code (here: https://github.com/grpc/grpc-java/blob/872b8c8792dcae9e29e79eba6fd7d21ac3049a51/netty/src/main/java/io/grpc/netty/NettyServerHandler.java#L607).


vietj:
I cannot see any test at the moment

vietj:
I still don't think that it is correct to do like this, cancel has a very precise meaning.

vietj:
for which reason envoy sends a reset with no error code ?

anavarr:
yeah i agree but have you looked at the grpc-java implementation? they basically treat all errors with a cancellation when receiving RST_STREAM

vietj:
what about other language / runtimes (.NET, etc..) ?

it would be good to know.

PR #371, opened 2026-09-16
Add gRPC over HTTP/3 support

### LINKED ISSUE #233: Add http3 support to grpc
### Describe the feature

We addition of http3 support to Vert.x core we should be able to move forward and adding support for grpc over http3. The standard already seems to specify that as per grpc proposal  https://github.com/grpc/proposal/blob/master/G2-http3-protocol.md I think more investigation should be done before hand for more sources and example implementations.

### Contribution

_No response_

arnabnandy7:
Motivation:

Add gRPC over HTTP/3 support following the protocol defined in the [gRPC HTTP/3 proposal](https://github.com/grpc/proposal/blob/master/G2-http3-protocol.md).

This change:

- Adds HTTP/3 as a supported native gRPC server protocol.
- Detects native gRPC requests received over HTTP/3.
- Allows clients configured with `HttpVersion.HTTP_3` to retain that transport instead of forcing HTTP/2.
- Reuses the existing gRPC framing implementation because gRPC request, response, DATA framing, and content-type semantics are unchanged over HTTP/3.
- Adds an end-to-end HTTP/3 unary-call test using TLS.
- Documents HTTP/3 client and server configuration requirements.

HTTP/2 remains the default client transport. HTTP/3 must be explicitly configured and requires TLS.

Fixes https://github.com/eclipse-vertx/vertx-grpc/issues/233

Conformance:

I have signed the Eclipse Contributor Agreement as explained in [CONTRIBUTING.md](https://github.com/eclipse-vertx/vert.x/blob/master/CONTRIBUTING.md).

The changes adhere to the [Vert.x code style guidelines](https://github.com/vert-x3/wiki/wiki/Vert.x-code-style-guidelines).


vietj:
I think that this does not handle the HTTP/3 specific error code changes between HTTP/2 and HTTP/3

arnabnandy7:
> I think that this does not handle the HTTP/3 specific error code changes between HTTP/2 and HTTP/3

@vietj thanks you're right. The normal request/response flow was using HTTP/3, but stream resets were still emitted and interpreted using HTTP/2 error codes.

working to update the reset handling to select the correct mapping based on the HTTP version, covering both client and server paths and tests for the HTTP/3 mappings defined by the G2 proposal.

PR #372, opened 2026-09-16
Extend event-bus gRPC response scope for streaming calls (#365)

### LINKED ISSUE #365: Consider extending the event-bus grpc response scope
### Describe the feature

The implementation of the gRPC protocol for the event-bus transport relies on an immediate response from the server. However this implies that tracing will be perceived differently for server streaming: the tracing span ends when the stream actually begins.

The protocol could be modified to accommodate this and let the server sends the response when streaming is finished. The consequences are:

- the event bus request might require an extended timeout (currently 30 seconds) for streaming
- the client must be ready to receive server frames before sending the request to avoid races
- the server to client frame can be replaced by a cancel response, very much like the HTTP/2 transport
- streaming client can use the response body instead of using a stream like a unary
- *however* there is no practical way for the client to ping the server, this needs another solution, e.g. rely on an ack frame to send the server details

### Contribution

_No response_

arnabnandy7:
Motivation:

The event-bus gRPC transport previously completed the initial event-bus request as soon as a streaming call was established. Consequently, distributed tracing recorded only stream setup rather than the complete lifetime of the call.

This change extends the event-bus response scope until the gRPC stream completes:

- Introduces an `Ack` transport frame for exchanging server endpoint details without completing the initial request.
- Registers the client endpoint before sending the request to avoid races with server frames.
- Completes or fails the event-bus request only when the stream finishes.
- Returns client-streaming unary responses through the final event-bus reply body.
- Maps final gRPC status and trailers to the event-bus reply.
- Adds a configurable streaming request timeout, defaulting to five minutes.
- Preserves flow-control ordering by waiting for queued response frames before completing the request.
- Validates malformed acknowledgement frames and fails affected calls immediately.
- Updates transport interception and documentation for the revised protocol.

Closes #365 

Validation:

- Focused regression suite: 43 tests passed.
- Complete `vertx-grpc-eventbus` suite: 91 tests passed.
- The `vertx-grpc-eventbus` module also passed in the repository-wide reactor build.
- The repository-wide build later failed in the unrelated `vertx-grpc-it` module because `io.vertx.codegen.format.LowerCamelCase` was unavailable during protobuf test-source generation.

Conformance:

I have signed the Eclipse Contributor Agreement as explained in [CONTRIBUTING.md](https://github.com/eclipse/vert.x/blob/master/CONTRIBUTING.md).

The changes adhere to the [Vert.x code style guidelines](https://github.com/vert-x3/wiki/wiki/Vert.x-code-style-guidelines).


vietj:
actually that was under consideration but is dismissed for the moment.

arnabnandy7:
> actually that was under consideration but is dismissed for the moment.

@vietj converted to draft, in future if this was considered back we can start from this point. 🤔

vietj:
sure thanks, actually we might in the future implement streaming in the event-bus itself, because this idea only covers the response and the request remains just the event bus message request, so it only fit 1/2 of the purpose.

in addition the event bus reply timeout, is really an issue.

vietj:
so in other words it's a false good idea

arnabnandy7:
> so in other words it's a false good idea

gotcha, can refactor when needed

vietj:
I think this branch will linger for some time, can you close this PR and keep your branch in your repo and link it in the issue instead ?

